PR #1, opened 2019-07-01
Completed Day1 Exercise

gauravj09:


gauravj09:
I will update this with a more efficient solution after finishing the second problem

ashmeetdeol19:
This solutions seems to work fine.
Go for optimization.

PR #5, opened 2019-07-01
Completed Day1 - Please review.

Shazmaan:


ashmeetdeol19:
Well Tried, but still you need to think more about this.

PR #6, opened 2019-07-01
Day 1 - Review 

Hariae:
This pull request is not a complete one, completed a basic working code. Need to work more on the problems.


SpecialStack.java
```diff
@@ -0,0 +1,71 @@
...
+
+    public int pop(){
+        return stack.pop();
+    }
+
+    public int peek(){
+        return stack.peek();
+    }
+
+    public int getMin(){
```
ashmeetdeol19: Will this work if you pop top few elements out of the stack ?
Will this function still give minimum value. ?
Think about it.

Hariae: I did not run my code against leetcode test cases so ended up missing many edge cases. Made changes to both problems based on inputs from sessions. 

https://github.com/super30admin/Day1/pull/6/commits/4346bce898f79580bd5a988f1c6ceddbc5303551

MyHashMap.java
```diff
@@ -0,0 +1,122 @@
...
+        System.out.println(myHashMap.get(23));
+        myHashMap.remove(13);
+        System.out.println(myHashMap.get(2));
+        myHashMap.remove(2);
+        System.out.println(myHashMap.get(2));
+        System.out.println(myHashMap.get(13));
+    }
+}
+
+/**
```
super30admin: Thats a nice code Hari.
Go for the next two problems.

PR #8, opened 2019-07-01
Completed Day1 - Please review

bharatreddypothireddy:
Used Internal stack for push,pop,isFull,isEmpty functions . Implemented getMin() in O(1) time using an extra stack .

### DIFF
--- a/NewHashMap.java
+++ b/NewHashMap.java
@@ -1,4 +1,4 @@
-/keys should be positive
+//keys should be positive
 public class NewHashMap {
     boolean[] keyArr = new boolean[1000000];
     int[] valueArr = new int[100000];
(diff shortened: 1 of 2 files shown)


NewHashMap.java
```diff
@@ -0,0 +1,31 @@
...
+
+    public static void main(String[] args) {
+        NewHashMap hm = new NewHashMap();
+        hm.put(3,33);
+        System.out.println(hm.get(3));
+        hm.remove(3);
+        System.out.println(hm.get(3));
+
+    }
+}
```
ashmeetdeol19: This seems to be more like a brute force solution.

bharatreddypothireddy: Yes , Will try to implement using linked lists .

bharatreddypothireddy:
Implemented using ListNodes

NewHashMap.java
bharatreddypothireddy: Implemented

PR #11, opened 2019-07-02
Hashmap and special Stack implementation completed-please review

PremalSamale:


HashMapImplementation.java
```diff
@@ -0,0 +1,55 @@
...
+  { 
+	  HashMap map=  new HashMap();
+      map.put(1, 10); 
+      map.put(2, 20); 
+      map.put(3, 30); 
+      map.get(1);
+      map.remove(1);
+  
+  } 
+}
```
ashmeetdeol19: This seems to be working but it is not the correct solution as you are using lot of space.
You need to  think about an optimal solution.

HashMap_SingleLinkedList.java
```diff
@@ -0,0 +1,160 @@
...
+	public static void main(String args[]) {
+		HashMap_Sll map = new HashMap_Sll();
+		map.put(1, 10);
+		map.put(2, 20);
+		map.put(3, 30);
+		map.get(1);
+		map.remove(1);
+
+	}
+}
```
ashmeetdeol19: Very well done with this solution.
If you are starting with size 16 then you also need to know the concept of rehashing.

MainSpecialStack.java
```diff
@@ -0,0 +1,69 @@
...
+		SpecialStack stack = new SpecialStack();
+		 stack.push(10);
+		 stack.push(11);
+		 stack.push(12);
+		 stack.getMin();
+		 stack.push(7);
+		 stack.pop();
+		 stack.getMin();
+}
+}
```
ashmeetdeol19: Nice job !!
The way you commented in between code is appreciable.

PR #13, opened 2019-07-02
Day 1 Problem 1 Completed the Hashmap - Please review.

lavanganji:


HashMap.java
```diff
@@ -0,0 +1,72 @@
...
+		hashMap.put(2, 2);         
+		System.out.println(hashMap.get(1));            // returns 1
+		System.out.println(hashMap.get(3));            // returns -1 (not found)
+		hashMap.put(2, 1);         					   // update the existing value
+		System.out.println(hashMap.get(2));            // returns 1 
+		hashMap.remove(2);          			       // remove the mapping for 2
+		System.out.println(hashMap.get(2));            // returns -1 (not found) 
+	}
+
+}
```
ashmeetdeol19: Doesn't this seem to be brute force solution ?

lavanganji: i did not get your question, what i mean to say the brute mean, i am using the two arrays. here to get the Hash map implement,  There are other way like implementing it with linked list. also. but i need to try that one.

PR #14, opened 2019-07-30
Completed HashMap and SpecialStack : Please Review

muditabysani:


muditabysani:
The complexity would be O(n) because I have to search if the key exists in the list.

jaspindersingh83:
Special stack push and pop are O(n)

muditabysani:
In python, for Append[1] operation average case complexity is O(1) and amortized worst case complexity is O(1).

MyHashMap.py
```diff
@@ -1,38 +1,66 @@
...
-        else:
-            return -1
+	# Returns the value to which the specified key is mapped, or -1 if this map contains no mapping for the key
+	def get(self, key):
+		i = self.find_index(key)
+		if self.Hashlist[i] == None:
+			return -1
+		prev_node = self.find_node(self.Hashlist[i], key)
+		# why should we check for this condition again because node.next would never be none if self.Hashlist[i] is not none
+		# And the length of our horizontal list is either zero or minimum two nodes right
```
super30admin: we need to check this condition because :
lets see what we are actually doing from the pur function 
we are first finding the index for the key we need to insert lets say the key with us is 2 now we went to the index 2 of our array if we find any other element already over there we are moving furthur in our nodes and inserting it at a place where there was null before
so now when i searched for that element i first checked the place where it was inserted if we find a null there that clearly means that key doesnt exist in our hashmap else it would be there
but if its not null over there i checked for nodes connected to it and traversed till i either got the node or i got null
in case i got null it means that key is not in our array so we again should check for this prev_node.next==null condition
and if its not null it means we have got the key and so we will return the value

PR #15, opened 2019-07-30
[Sweta Kumari] Adding Problem2 of Design-1

SwetaK951:


SpecialStack.java
```diff
@@ -0,0 +1,73 @@
...
+    private int min = Integer.MAX_VALUE;
+    private Stack<Element> specialStack = new Stack<Element>();
+
+    class Element{
+        private int value;
+        private int currMin;
+
+        Element(int value, int currMin){
+            this.value = value;
+            this.currMin = min;
```
kaurramanpreet: you mean
this.currMin=currMin

SwetaK951:
Yes, It will be currMin. I have added a new approach to this problem.

PR #20, opened 2019-08-19
explanation in the same file below code

PATIL1695:


minstack.py
```diff
@@ -0,0 +1,47 @@
...
+    def __init__(self):
+        self.s = []
+
+    def push(self,x):
+        cur_min = self.getMin()
+        if cur_min == None or x < cur_min:
+            cur_min = x
+        self.s.append((x,cur_min))
+
+    def pop(self):
```
kaurramanpreet: should you pop by checking some of the conditions or you should only pop the topmost element ?

PATIL1695:
Yeah I need to check if the stack is empty or not before popping. If it's
not empty then only I should pop.

Warm regards-
Vishwanath Patil

kaurramanpreet:
Did you attend today's session

kaurramanpreet:
Try the question again submit on leetcode and check 
Then submit the pr again

PATIL1695:
Yeah I did.

Warm regards-
Vishwanath Patil

PATIL1695:
Okay sure.

Warm regards-
Vishwanath Patil

PR #39, opened 2019-10-09
Design 1 Assignment

88sanjay:


sehaj-grover (commented):
Sanjay implement hashmap using hashindex and handle collisions in that scenerio because that is what the interviewer wants

88sanjay:
> Sanjay implement hashmap using hashindex and handle collisions in that scenerio because that is what the interviewer wants

Sure. I can do that. Though I think this is more optimal from a space.  I have implemented the double hashing solution in the hashset assignment

PR #47, opened 2019-11-19
Completed Problem2 of Design-1 - Please review

noolvikiran24:


noolvikiran24:
> Good job, try two stack approach also in MinStack .Try to solve Design HashMap problem as well.

Sure, I have submitted the Hashmap problem and I will submit the problem 1 using 2 stacks as well.

PR #59, opened 2019-11-24
Finished Design hash map

sudhasr:


DesignHashMap.py
```diff
@@ -0,0 +1,96 @@
+"""Could you please check the time and space complexities?"""
+
+#Accepted on leetcode
+
+#Time complexity for creating Hash Map - O(N)
+#Space omplexity - O(1) or O(k)
```
neha79091: Space complexity will not be constant as you are having n elements inside it

DesignHashMap.py
```diff
@@ -0,0 +1,96 @@
...
+        if cur.pair[0] == key:
+            self.nodes[index]=cur.next
+        else:
+            cur = cur.next
+            while cur:
+                if cur.pair[0] == key:
+                    prev.next =cur.next
+                    break
+                else:
+                    cur,prev = cur.next,prev.next
```
neha79091: instead of writing iteration again and again,you can write a findnode function.

PR #62, opened 2019-12-23
Special Stack - Please Review

BhushanVasisht:
Hello. 
I am kinda confused here. The problem statement in github abd leetcode do not match.
I am implementing a stack according to the reqirement given on leetcode page. But the problem statement says to not use arrays/lists and to use only standard Stack. Could you please confirm.

Thanks

### DIFF
--- a/HashMap.java
+++ b/HashMap.java
@@ -1,0 +1,63 @@
+// Time Complexity :
+//      put() - O(1)
+//		get() - O()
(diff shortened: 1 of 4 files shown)


kaurramanpreet:
well done bushan
you have to complete the minstack problem using 2 stacks and 1 stack solution!!
try them again after attending the session!!

PR #66, opened 2019-12-23
Uploading Solutions 1 and 2

mdave1701:
Completed Solutions to problems 1 and 2 in python.

### DIFF
--- a/MinStack.py
+++ b/MinStack.py
@@ -1,0 +1,85 @@
+"""
+Time Complexity: O(1)
+Space Complexity: O(n)
+Did this code run successfully on leetcode: No, this problem was slightly different than the one on leetcode.
+
+Problem: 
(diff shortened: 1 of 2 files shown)


MinStack.py
```diff
@@ -0,0 +1,85 @@
...
+Space Complexity: O(n)
+Did this code run successfully on leetcode: No, this problem was slightly different than the one on leetcode.
+
+Problem: 
+Design a SplStack that supports all the stack operations push, pop, isEmpty, isFull and an operation getMin() 
+which should return minimum element from the SplStack. All these operations must be O(1). To implement this, 
+you should only use standard Stack data structure.
+"""
+
+class MinStack:
```
kaurramanpreet: try this problem using 2 stack and 1 stack as well!!

PR #86, opened 2020-01-02
S30 Big N Problem #2 {Medium}

kamaleshanantha:
Design a SplStack that supports all the stack operations push, pop, isEmpty, isFull and an operation getMin() which should return minimum element from the SplStack. All these operations must be O(1). To implement this, you should only use standard Stack data structure.


kamaleshanantha:
> Good, Try other problem and try to solve this problem using one stack also.

Uploaded the solution for Design Hashmap. Kindly review. Thanks

PR #91, opened 2020-02-04
Submit 1 and 2

siddheshgupte:


siddheshgupte:
The hashmap implementation is the chaining solution! First array is of size 10 with the "hashing function" as %10 and then at each position of the array is a python list with the tuple (original_key, value)

siddheshgupte:
Added 2 stack approach

PR #112, opened 2020-02-22
Completed Design-1

SweetySojrani:


Problem2.java
```diff
@@ -0,0 +1,53 @@
+// Using 1 stack
+// Time Complexity - O(n) where n is the number of operations
+//Space Complexity - O(n) where n is the number of elements to push on stack
```
neha79091: think about the time complexity again

neha79091:
Try doing two stack solution also

PR #113, opened 2020-03-10
Design 1

vaidyapawan:
706. Design HashMap
155. Min Stack


devashishnyati:
Good job Pawan!
Please add all the 4 things provided in the sample.java file: time, space, successful on leetcode and challenges.
Also, think about handling collisions in the hashmap.

PR #114, opened 2020-03-10
Design-1

AishwaryaSharma27:
Please review. 
(Both questions in same file)

### DIFF
--- a/Sample.java
+++ b/Sample.java
@@ -1,7 +1,78 @@
-// Time Complexity :
-// Space Complexity :
-// Did this code successfully run on Leetcode :
-// Any problem you faced while coding this :
+// Time Complexity : Push - O(1), Pop -> O(1), Min -> O(1)
+// Space Complexity : O(n)
+// Did this code successfully run on Leetcode : Yes
+// Any problem you faced while coding this : not sure about the space complexity as this is what we have to return so will this stack space be included in the space complexity.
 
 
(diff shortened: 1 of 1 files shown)


Sample.java
```diff
@@ -1,7 +1,78 @@
-// Time Complexity :
-// Space Complexity :
-// Did this code successfully run on Leetcode :
-// Any problem you faced while coding this :
+// Time Complexity : Push - O(1), Pop -> O(1), Min -> O(1)
+// Space Complexity : O(n)
+// Did this code successfully run on Leetcode : Yes
+// Any problem you faced while coding this : not sure about the space complexity as this is what we have to return so will this stack space be included in the space complexity.
```
devashishnyati: You can ask this question to the interviewer. Generally, if you are not returning the data structure then it is taking extra space. So yes it should be O(n) here.

Sample.java
```diff
@@ -1,7 +1,78 @@
...
+        return stack.peek()[0];
+    }
+    
+    public int getMin() {
+        return stack.peek()[1];
+    }
+}
+
+
+// Time Complexity : Add - O(1), Remove -> O(1), Get -> O(1)
```
devashishnyati: Try to use hashing here

devashishnyati:
Good job Aishwarya!
Please submit different files for different problems.

PR #115, opened 2020-03-11
Completed Design I problems

arcsoftech:


devashishnyati:
Good job Arihant!
Try to implement hashmap using double hashing also to get better time complexity.

PR #116, opened 2020-03-11
Design 1 problems

Sandersd:
706. Design HashMap
155. Min Stack

### DIFF
--- a/hashmap.js
+++ b/hashmap.js
@@ -1,0 +1,50 @@
+// Time Complexity : O(1)
+// Space Complexity : O(n)
+// Did this code successfully run on Leetcode : Yes
+// Any problem you faced while coding this :
+
+/**
+ * Initialize your data structure here.
+ */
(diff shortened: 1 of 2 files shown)


minStack.js
```diff
@@ -0,0 +1,54 @@
+// Time Complexity : O(1)
+// Space Complexity : O(n)
+// Did this code successfully run on Leetcode : Yes
+// Any problem you faced while coding this: Is sorting allowed?
```
devashishnyati: No, sorting is not allowed. If you are using sorting, the time complexity changes to O(nlogn).

hashmap.js
```diff
@@ -0,0 +1,50 @@
+// Time Complexity : O(1)
```
devashishnyati: Try it with using hashing

devashishnyati:
Good job Dylan!!
Please try the problems by using hashing and not using sorting.

PR #119, opened 2020-03-11

KBangera:
Need to think of a better approach for the Spl Minstack problem

### DIFF
--- a/Sample.java
+++ b/Sample.java
@@ -1,7 +1,86 @@
-// Time Complexity :
-// Space Complexity :
-// Did this code successfully run on Leetcode :
-// Any problem you faced while coding this :
+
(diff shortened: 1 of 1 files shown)


Sample.java
```diff
@@ -1,7 +1,86 @@
-// Time Complexity :
-// Space Complexity :
-// Did this code successfully run on Leetcode :
-// Any problem you faced while coding this :
+
+// Time Complexity :O(1)
+// Space Complexity :O(n)
```
devashishnyati: Try to optimize the space by using hashing

devashishnyati:
Good job Karthik!
Please submit 2 files for 2 problems.
Also, try to implement hashmap with linear chaining and double hashing!

PR #121, opened 2020-03-11
Added my Solutions

AroopN:
**Problem 1**
The solution for Designing Hashmap did not run within time limits. I used separate chaining for collision avoidance.

**Problem 2**
Implemented MinStack using stack of lists with O(1) time complexity.


devashishnyati:
Good job Aroop!
Try to implement hashmap with double hashing also.

PR #123, opened 2020-03-12
Design-1 Min Stack Completed

AMAbhi:
Please Review - Min Stack


devashishnyati:
Good job Abhishek!
Please add the 4 things at the top here also.
Try to implement this with one stack.
Also, submit the other problems.

PR #129, opened 2020-04-20
Answer to question 1 and 2 in Sample.java - Please review

Abinashi7:
Not able to understand the runtime complexity of the solution provided on Leetcode

### DIFF
--- a/Sample.java
+++ b/Sample.java
@@ -1,7 +1,122 @@
-// Time Complexity :
-// Space Complexity :
-// Did this code successfully run on Leetcode :
-// Any problem you faced while coding this :
+// Time Complexity : O(1)
+// Space Complexity :0(n)
+// Did this code successfully run on Leetcode : Yes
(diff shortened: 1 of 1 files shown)


Sample.java
```diff
@@ -1,7 +1,122 @@
...
 
 // Your code here along with comments explaining your approach
+// Create an array of upperbound 10000. If a key is 2 with value 5. It will be stored like this [][][5][]...[n]
+class MyHashMap {
+    int[] store;
+    boolean[] seen;
+    /** Initialize your data structure here. */
+    public MyHashMap() {
+
+        store = new int[100000];
```
super30admin, line 15: We have committed extra space of 2 million bits over here always.

Sample.java
```diff
@@ -1,7 +1,122 @@
...
+
+    public class pair {
+        int min;
+        int val;
+
+        pair(int val, int min) {
+            this.val = val;
+            this.min=min;
+        }
+    }
```
super30admin, line 74: Can we do it without need of another Data Structure like pair ?

super30admin:
Good Try Abhinashi. We will be learning better solutions in the class

PR #130, opened 2020-04-20
Completed Design-1: Please Review

anikpait:


Exercise_2_MinStack.java
```diff
@@ -0,0 +1,49 @@
+// Time Complexity : push(), pop(), top() , getMin()   -> O(1) for each element
+// Space Complexity : O(n) where n is size of the stack and minStack (temporary stack used)
+// Did this code successfully run on Leetcode : Yes
+/* Any problem you faced while coding this : None
+ */
+import java.util.Stack;
+class MinStack {
+    
+    Stack<Integer> stack;
+    Stack<Integer> minStack;
```
super30admin, line 10: Good Can we do it using single stack?

Exercise_1_HashMap.java
```diff
@@ -0,0 +1,45 @@
...
+ */
+import java.util.Arrays;
+class MyHashMap {
+
+	// Your code here along with comments explaining your approach
+
+    int[] store;
+    /** Initialize your data structure here. */
+    public MyHashMap() {
+        store = new int[1000001]; // Max number of key-value pairs allowed in the HashMap, cant exceed it.
```
super30admin, line 16: Not a good way. We would learn why?

super30admin:
Good Job Anik

Exercise_2_MinStack.java
anikpait: I think, we would require one stack to return the results and another temporary storage to keep a track of the minimum we have till now. Using a single stack can only fulfill one purpose at a time. If it keeps a track of minimum, it would need to ignore the other elements and our push and pop operations would be affected. It will be interesting to see if it is feasible with one stack but I think we would require a temp storage.

super30admin: We will see a single stack solution as well tomorrow

PR #136, opened 2020-04-21
Completed Problems of Design-1. Please Review

sbsreedh:


minStack.py
```diff
@@ -0,0 +1,48 @@
+#Time Complexity : pop:O(1), top:O(1) , push:O(1), getMin: O(1)
+# Space Complexity :O(1)
```
super30admin, line 2: Think again? We are not using constant/no space....

minStack.py
```diff
@@ -0,0 +1,48 @@
...
+        """
+        self.stack=list() #initialize stack using list
+
+    def push(self, x: int) -> None:
+    #obtain element with minimum value
+        minVal=self.getMin()
+        #check if the minimum value exits and the number to be pushed on stack is lesser than the minimum value obtained.
+        if minVal is None or x< minVal:
+            minVal=x# if x is lesser then min value, we update the min value
+        self.stack.append((x,minVal))# appens the list with the min value at index 1.
```
super30admin, line 24: This is O(n) operation

minStack.py
```diff
@@ -0,0 +1,48 @@
+#Time Complexity : pop:O(1), top:O(1) , push:O(1), getMin: O(1)
```
super30admin, line 1: Time complexity is not O(1) as per code implemented in the problem

myHashMap.py
```diff
@@ -0,0 +1,84 @@
...
+        """
+        Returns the value to which the specified key is mapped, or -1 if this map contains no mapping for the key
+        """
+        idx=key%self.size # get the index value 
+        # kv_pair(key,value)
+        
+        if self.map[idx] ==-1:
+            return -1u
+        # if the cell is not None, we iterate through the list to find if key exists and if so we return its value.
+        for kv_pair in self.map[idx]:
```
super30admin, line 49: If we have to iterate over the whole map every time then it beats the purpose of hashMap. We could have easily converted this solution in O(1) always by taking two separate hash functions

myHashMap.py
```diff
@@ -0,0 +1,84 @@
+#Time Complexity : O(n)- Here n is 1000, as I have defined size as 1000
```
super30admin, line 1: This is wrong way of thinking about Time Complexity

super30admin:
Good Job. Please review inline comments. Ask questions in class if not clear

minStack.py
sbsreedh: The TC of push() is O(n)? Since I am calling getmin() everytime I am calling push() function. Please correct if I am wrong.

myHashMap.py
sbsreedh: The TC is O(n), where n will be the number of elements present in stack and not the size defined in the code. Is that right?

PR #144, opened 2020-04-23
[Design-1] Review Please

yosnehal:


Sample.java
```diff
@@ -1,7 +1,132 @@
...
+/**
+ * Your MyHashMap object will be instantiated and called as such:
+ * MyHashMap obj = new MyHashMap();
+ * obj.put(key,value);
+ * int param_2 = obj.get(key);
+ * obj.remove(key);
+ */
+
+// Time Complexity : Each of the operation will be O(1)
+// Space Complexity : Each of the operation will be O(1)
```
iivaishnavii, line 85: Since we use a stack data structure, the time complexity will be O(n).

iivaishnavii:
Good Job, Snehal! Keep an eye out for additional data structures while calculating space complexities.

Sample.java
ssgandham: Thanks @iivaishnavi for letting em know. You mean O(n) for all the operations or for single operation.

PR #150, opened 2020-04-27
completed

namrataKasar:


Sample.java
```diff
@@ -1,7 +1,98 @@
...
+        }
+    }
+}
+
+
+// Time Complexity : O(1) for push, pop, top and getMin
+// Space Complexity : O(N) which is stack size
+// Did this code successfully run on Leetcode : Yes
+// Any problem you faced while coding this : No
+
```
iivaishnavii, line 59: Interesting to see the Stack<int[]> approach instead of using two stacks / pushing the elements twice into stack

Sample.java
```diff
@@ -1,7 +1,98 @@
-// Time Complexity :
-// Space Complexity :
-// Did this code successfully run on Leetcode :
-// Any problem you faced while coding this :
+// Time Complexity : O(N) as it uses contains method
+// Space Complexity : O(N)
+// Did this code successfully run on Leetcode : Yes
+// Any problem you faced while coding this : Time complexity is not O(1) for all operations
```
iivaishnavii, line 4: Good try! After class, do submit the O(1) approach too.

iivaishnavii:
Good Job, Namrata! Nice to see you proactively submitting PRs.

PR #156, opened 2020-05-25
HashMap- No Collision Control

Keerti995:
HashMap - No Collision Control

### DIFF
--- a/Sample.java
+++ b/Sample.java
@@ -1,7 +1,0 @@
-// Time Complexity :
-// Space Complexity :
-// Did this code successfully run on Leetcode :
-// Any problem you faced while coding this :
-
-
-// Your code here along with comments explaining your approach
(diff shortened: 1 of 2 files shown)


src/HashMap_simple.java
```diff
@@ -0,0 +1,94 @@
...
+ */
+public class HashMap_simple {
+    int[] values;                       //stores the values with key being the array index
+    int capacity;                       // default size of the array
+    int size;                           // filled array size
+
+    HashMap_simple(){
+        capacity = 16;                  //Default value used by Java too
+        values = new int[capacity];
+        fillArray(values);              //filling the values of the array with dummy value
```
chiragshahi, line 21: Can use inbuilt function:
Arrays.fill(values, Integer.MIN_VALUE)

chiragshahi:
Good job Keerti! 
Write time, space complexities 
See inline comments
Try doing hash map using an array of linked lists as well

PR #157, opened 2020-05-25
Completed Sprint Design-1 - Please Review

ktkrutik:


MinStack.java
```diff
@@ -0,0 +1,57 @@
+// Time Complexity : O(1)
+// Space Complexity : O(n)
```
chiragshahi, line 2: where n is ?
define n as well

chiragshahi:
Great job with Q1
Attempt Q2 using an array of linked lists since the array approach is suboptimal

MinStack.java
ktkrutik: Where n is size of input

PR #163, opened 2020-05-25
Completed Design-1 Please review, No Collision control for Hash Map

guravashish:
Completed Design-1 Please review
No Collision control for Hash Map

### DIFF
--- a/HashMap.java
+++ b/HashMap.java
@@ -1,0 +1,52 @@
+ import java.util.List;
+
+import sun.net.www.content.text.plain;
+
+import java.util.LinkedList;
+// Time Complexity : O(1)
+// Space Complexity : O(n)
+// Did this code successfully run on Leetcode : Yes
+// Any problem you faced while coding this : No Collision Control
+
+// Your code here along with comments explaining your approach
+public class HashMap {
+    List<Integer> listKey;
+    List<Integer> listValue;
+
+
+    //constructor
+    public HashMap(){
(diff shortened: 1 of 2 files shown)


MinStack.java
```diff
@@ -0,0 +1,83 @@
+import java.util.Stack;
+
+// Time Complexity : O(1)
+// Space Complexity : O(n)
+// Did this code successfully run on Leetcode : Yes
+// Any problem you faced while coding this : Did not get the logic to solve isFull()
```
chiragshahi, line 6: Are we using isFull() anywhere?

chiragshahi:
Good job writing clean code with comments, time space complexties 
See inline comment for designing hashmap

MinStack.java
guravashish: No, but as it was asked the problem statement that is why I was wondering "Design a Data Structure SpecialStack that supports all the stack operations like push(), pop(), isEmpty(), isFull()"

HashMap.java
```diff
@@ -0,0 +1,52 @@
+ import java.util.List;
+
+import sun.net.www.content.text.plain;
+
+import java.util.LinkedList;
+// Time Complexity : O(1)
+// Space Complexity : O(n)
+// Did this code successfully run on Leetcode : Yes
+// Any problem you faced while coding this : No Collision Control
```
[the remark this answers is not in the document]
guravashish: got it that thankyou!

PR #166, opened 2020-05-25
Completed Design 1 Please Review

sridhar-anish-001430789:


MyHashMap.java
```diff
@@ -0,0 +1,40 @@
+
+// Time Complexity : O(1)
+// Space Complexity :O(n)
```
Lohith-Aswathappa, line 3: If you are using the constant space, space complexity would be O(1)

Lohith-Aswathappa: Try to the solution taught in class

Lohith-Aswathappa:
Nice try, Anish!
Try the solutions taught in class.

PR #168, opened 2020-05-25
Hashmap and MinStack Implementations

tramatejaswini:


Hashmap_implementation.py
```diff
@@ -0,0 +1,84 @@
+// Time Complexity : O(1) Best , O(N/1000) worst time complexity
+// Space Complexity : O(N) where N is the number of items in the dictionary
+// Did this code successfully run on Leetcode : Yes
+// Any problem you faced while coding this : No, but getting this idea is very different from other problems,
```
Lohith-Aswathappa, line 4: Nice try Tejaswini, Since you have coded it now, try to code the solution again and see if you encounter the same problem

Lohith-Aswathappa:
Good work, Tejaswini!

Hashmap_implementation.py
tramatejaswini: Thanks, Lohith. I tried the solution with double hashing as well which is more optimized than the LinkedList. I did not encounter any problem.

PR #169, opened 2020-05-25
Design1 sprint complete - Please review

osamamohamedsoliman:


Problem_1.py
```diff
@@ -0,0 +1,87 @@
+# Time Complexity :O(1)
```
Lohith-Aswathappa, line 1: Please check the time complexity

Lohith-Aswathappa:
Good job Osama! 
Please review the inline comments

Problem_1.py
osamamohamedsoliman: Resolved :)

PR #171, opened 2020-05-25
Completed Design-1, kindly review and share feedback.

Sumesh-H:


MyHashMap.java
```diff
@@ -0,0 +1,85 @@
+// Time Complexity : finding element is O(1)
```
chiragshahi, line 1: // Time Complexity : O(n) where n is the length of the LL formed at any index of the array
// O(1) for finding in array * O(n) for finding in LL = O(n)

chiragshahi:
Excellent work Sumesh! Be careful writing correct time, space complexities

PR #173, opened 2020-05-25
Design 1 Completed in Java - Please review

Pradhisha:


Design_HashMap.java
```diff
@@ -0,0 +1,39 @@
+// Time Complexity : O(1)
```
Lohith-Aswathappa, line 1: Nice try Pradhisha!
Try the solution taught in class.

Lohith-Aswathappa:
Good work, Pradhisha!
Please review inline comments.

PR #174, opened 2020-05-25
Completed Hashmap and MinStack

GahanJagadeesh:


MinStack.java
```diff
@@ -0,0 +1,50 @@
+// Time Complexity : O(1)
+// Space Complexity :O(1)
```
Lohith-Aswathappa, line 2: Good job Gahan, Try the 2 stacks solutions as well.

MyHashMap.java
```diff
@@ -0,0 +1,45 @@
+
+// Time Complexity : O(n)
```
Lohith-Aswathappa, line 2: Since you are accessing the elements without traversing all the elements, the time complexity would be O(1)

Also, try the solution taught in class

Lohith-Aswathappa:
Good try Gahan!
 Check the inline comments.

PR #175, opened 2020-05-26
Completed Design-1 -Please Review

DineshNadar95:


MinStack.java
```diff
@@ -3,87 +3,53 @@
...
-            return 0;
-        } else {
-            return a[top];
-
-        }
+       return (int) minStack.peek();
 
     }
 
     public int getMin() {
```
chiragshahi, line 32: This defeats the purpose of the question. you cannot search thru the stack for the minimum element

MinStack.java
```diff
@@ -0,0 +1,93 @@
+// Time Complexity : O(n)
+// Space Complexity :O(n)
+// Did this code successfully run on Leetcode : Yes
+// Any problem you faced while coding this : is time complexity right for this problem i am a bit confused.
```
chiragshahi, line 4: Yes it is O(n) because you searched thru all the elements in the stack
The question asks to complete it in constant time

MyHashMap.java
```diff
@@ -0,0 +1,40 @@
+// Time Complexity : O(1)
+// Space Complexity :O(1)
+// Did this code successfully run on Leetcode : Yes
+// Any problem you faced while coding this : is time complexity right for this problem i am a bit confused.
```
chiragshahi, line 4: No, it should be O(n) since Array.fill uses O(n) time

chiragshahi:
Good job with hash map
Remove unwanted functions from the code 
Be careful to consider what algo or DS you're using before writing space time complexities

MyHashMap.java
DineshNadar95: Is the space complexity correct for this solution, even if I initialize  the array with the length of 1 billion the space complexity would still be O(1)?

PR #176, opened 2020-05-26
Design 1 done

AbhijeetDhillon:


MinStack.java
```diff
@@ -0,0 +1,51 @@
+// Time Complexity :O(n)
```
chiragshahi, line 1: This defeats the purpose of the question. you cannot search thru the stack for the minimum element

MinStack.java
```diff
@@ -0,0 +1,51 @@
+// Time Complexity :O(n)
+// Space Complexity :O(1)
```
chiragshahi, line 2: Space complexity is O(n) where n is the size of the stack

chiragshahi:
Good job with hash map
Be careful to consider what algo or DS you're using before writing space time complexities

PR #178, opened 2020-05-26

ashiranka:


MinStack.java
```diff
@@ -0,0 +1,43 @@
+// Time Complexity : O(n)
```
chiragshahi, line 1: This defeats the purpose of the question. you cannot search thru the stack for the minimum element

chiragshahi:
Good job with hash map 

PR #182, opened 2020-05-26
Completed Hashing & MinStack : Please Review

shantanu609:
Please Review

### DIFF
--- a/MinStack.py
+++ b/MinStack.py
@@ -1,0 +1,35 @@
+class MinStack:
+
+    def __init__(self):
+        """
+        initialize your data structure here.
+        """
+        self.stack = []
+
+    def push(self, x: int) -> None:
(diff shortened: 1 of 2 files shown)


MinStack.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def pop(self) -> None:
+        if len(self.stack) >0:
+            return self.stack.pop()
+
+    def top(self) -> int:
+        if len(self.stack) >0:
+            return self.stack[-1]
+
+    def getMin(self) -> int:
```
AbhiniveshP, line 20: This approach is not the optimal one as getMin() always takes O(n)

AbhiniveshP:
Kindly include a 3-step approach at the beginning describing briefly what your approach was, along with the Time and Space Complexities mentioned.

PR #184, opened 2020-05-26
Completed both problems, please review.

samarthjj:


AbhiniveshP:
Great job with the comments at every line of code. Kindly include a 3-step approach at the beginning describing briefly what your approach was, apart from the Time and Space Complexities mentioned.

samarthjj:
Thanks for the feedback, I'll definitely keep this in mind for future PRs

PR #186, opened 2020-05-27
Completed Design 1 in Python

pcvyas:


Lohith-Aswathappa:
Nice try, Priyansh!
Try the solutions taught in class.

PR #195, opened 2020-05-28
q-1 hashmap custom implementation

ashwin1:


Lohith-Aswathappa:
Good work, Ashwin. 
Try the minStack problem as well.

PR #198, opened 2020-05-29
HashMap and MinStack

PadmaPriyaCheruvu:


MyHashMap.java
```diff
@@ -0,0 +1,87 @@
+// Time Complexity : O(1) for get, put and remove
```
chiragshahi, line 1: // Time Complexity : O(n) where n is the length of the LL formed at any index of the array
// O(1) for finding in array * O(n) for finding in LL = O(n)
// Space Complexity : O(n) where n is the length of the LL formed at any index of the array
// O(1000) for array * O(n) for LL = O(n)

chiragshahi:
Great job! See inline comment for correct time space complexity

PR #207, opened 2020-07-07
Design HashMap completed in java

Shilpi10:
Please check

### DIFF
--- a/HashMapImplement.java
+++ b/HashMapImplement.java
@@ -1,0 +1,115 @@
+// Time Complexity : Amortized O(1) for put, get and remove
+// Space Complexity : O(n) where n is number of element in th hashmap
+// Did this code successfully run on Leetcode : Yes
+// Any problem you faced while coding this : for get(-8) i faced problem while handling it, but figured out the way
+
+
+public class HashMapImplement {
+    class ListNode
+    {
+        //creating structure for each listnode
+        int val;
+        int key;
+        ListNode next;
+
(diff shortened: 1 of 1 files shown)


Shilpi10:
Have done min stack also so pls check

HashMapImplement.java
```diff
@@ -0,0 +1,115 @@
...
+        }
+    }
+    // using array of linkedlist to implement hashmap
+    ListNode[] map;
+    public HashMapImplement() {
+        map = new ListNode[10000];
+    }
+    public void put(int key, int value) {
+        int index = getIndex(key); // gets the index
+        ListNode valNode = findElement(index, key); // goes to the index and finds key
```
samarthjj, line 28: Put will crash, if valNode is null.

HashMapImplement.java
```diff
@@ -0,0 +1,115 @@
...
+        if(valNode.next != null)
+        {
+            //if present the present node will now point to the next of its next
+            valNode.next = valNode.next.next;
+        }
+
+    }
+    public int getIndex(int key)
+    {
+        if(key < 0)// for out of bound case
```
samarthjj, line 64: A HashMap should be able to store any integers as keys.

samarthjj:
Good job Shilpi, Please review the inline comments. 
Please try the single stack implementation for MinStack.
What was the exact problem you faced for HashMap? Please text me on Slack.

PR #212, opened 2020-07-08
Added Implementation for Design-1 Problems

goelsarthak:
Please review the solutions of below problems:

1.  Design a Hash Map.
2. Implement Min Stack.

Thanks,
Sarthak

### DIFF
--- a/Problem2.java
+++ b/Problem2.java
@@ -1,0 +1,53 @@
+class MinStack {
+
+    /** initialize your data structure here. */
+
+    //Stack that gonna store all the elements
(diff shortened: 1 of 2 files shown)


Sumesh-H:
Good Job Sarthak! Kindly mention Time & Space complexities for both problems.

Problem2.java
```diff
@@ -0,0 +1,53 @@
+class MinStack {
```
Sumesh-H, line 1: Try doing it using one stack also.

PR #213, opened 2020-07-08
Solutions to both problems - please review

khushbooG9:


samarthjj:
Nice work Khushboo. Your implementation of MinStack is essentially how it's done using two stacks, keeping track of min for every element. Please do also try the single stack approach.

PR #217, opened 2020-07-08
Design 1 HashMap and Min Stack in Java || Please review

shashidhar-01:
Java

### DIFF
--- a/MinStack.java
+++ b/MinStack.java
@@ -1,0 +1,48 @@
+// Time Complexity : O(1)
+// Space Complexity : O(n)
+// Did this code successfully run on Leetcode : yes
+// problems : i was getting empty stack exception , was using min_stack.peek()>x in line 12's if condition
+import java.util.*;
+class MinStack {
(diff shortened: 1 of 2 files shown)


MinStack.java
```diff
@@ -0,0 +1,48 @@
+// Time Complexity : O(1)
+// Space Complexity : O(n)
+// Did this code successfully run on Leetcode : yes
+// problems : i was getting empty stack exception , was using min_stack.peek()>x in line 12's if condition
```
Sumesh-H, line 4: You need to check both x<= min as to maintain the previous.

MyHashMap.java
```diff
@@ -0,0 +1,77 @@
+// Time Complexity : O(1) //amortized
+// Space Complexity :
```
Sumesh-H: O(n) where n is the length of the LL formed at any index of the array.
O(10000) for array * O(n) for LL = O(n).

Sumesh-H:
Good Job Shashidhar! Kindly go through inline comments.

PR #219, opened 2020-07-08
Completed Design - 1 Hashmap and MinStack (modified it). Kindly review it !

sneharnair:


Problem-1_Design_Hashmap.py
```diff
@@ -0,0 +1,78 @@
+# Time Complexity : O(1) - get_ind, put, get and remove functions. O(n) - find_node function where n is the length of the linked list at that
```
samarthjj, line 1: put, get and remove all make an internal call to find_node, hence the TC should O(n), not O(1).

samarthjj:
Nice work Sneha, Please review the inline comment. Nicely commented throughout!
For minStack please add edge cases to prevent pop/top on operating on an empty stack.

PR #229, opened 2020-07-08
completed design-hashmap and min_stack. please Review. 

satishsolanki1990:


min_stack.py
```diff
@@ -0,0 +1,71 @@
...
+        initialize your data structure here.
+        """
+        self.stack = []
+        self.min=sys.maxsize
+
+    def push(self, x: int) -> None:
+        """
+        if incoming is less or equal to min than push(min) -> min=incoming -> push(incoming)
+        """
+        if x >= self.min:
```
samarthjj, line 16: This should be strictly greater (>) not >=

samarthjj:
Good work Satish, please review the inline comments.
Please comment your code and add the Time & Space complexity, refer to the sample.java file.

min_stack.py
satishsolanki1990: understood. Thanks

satishsolanki1990: desgin_hashmap shows error while submitting at Leetcode. It throws error when a large number of operations got performed.

PR #231, opened 2020-07-09
Please review - included one stack solution for minstack

khushbooG9:


minstack.py
```diff
@@ -0,0 +1,72 @@
...
+        self.stack.append(x)
+
+    def pop(self) -> None:
+        val = self.stack.pop()
+        if self.min==val:
+            newmin = self.stack.pop()
+            self.min = newmin
+
+    def top(self) -> int:
+        return self.stack[-1]
```
samarthjj, line 31: for pop and top, please add edge cases to prevent from operating on an empty stack.

samarthjj:
Good job Khushboo, please review the inline comments. Please make sure to add your approach as comments in 3-4 bullet points for every problem/approach.

PR #233, opened 2020-07-09
Create Hash Map Design - Please Review

firoz1905:
I tried to implement the way Dominic explained of not creating the dummy nodes in the constructor but creating them in the find element function.But I keep getting the error self is not defined for ListNode.

may be you can help , I don't understand why we haven't created the list where the keys without collisions need to be saved . Once we see the collisions then only we need to initiate the  linked list nodes and traverse through them.

Thank you  in advance for your help !! I am coming from non development background.

### DIFF
--- a/Hash Map Design
+++ b/Hash Map Design
@@ -1,0 +1,73 @@
+class MyHashMap:
+    
+    SIZE=1000
+    nodes=[ListNode(-1) for i in range(SIZE)] 
+    
+    def __init__(self,):
+        """
+        Initialize your data structure here.
+        """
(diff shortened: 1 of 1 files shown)


Sumesh-H:
> I tried to implement the way Dominic explained of not creating the dummy nodes in the constructor but creating them in the find element function.But I keep getting the error self is not defined for ListNode.
> 
> may be you can help , I don't understand why we haven't created the list where the keys without collisions need to be saved . Once we see the collisions then only we need to initiate the linked list nodes and traverse through them.
> 
> Thank you in advance for your help !! I am coming from non development background.

Hi firoz, 
Kindly message me on slack. Let's work on it together.

PR #238, opened 2020-07-13
completed-please review

JaneJianLee:


solution2.py
```diff
@@ -0,0 +1,45 @@
...
+            self.s1.append(self.min)
+            self.min = x
+        self.s1.append(x)
+
+    def pop(self) -> None:
+        val = self.s1.pop()
+        if val == self.min:
+            self.min = self.s1.pop()
+
+    def top(self) -> int:
```
samarthjj, line 34: pop and top will crash if called on an empty stack, please add error guarding if statements to prevent from operating on an empty stack.

samarthjj:
Good work Jane, Review the inline comments. Please add comments explaining your approach.
Please complete the other problem **Design HashMap** as well.

PR #240, opened 2020-07-14
Please review

EnnaSachdeva:
The codes run perfectly for simple cases, but fails in Leetcode, not able to figure out the issue. Please review.

### DIFF
--- a/hashmap_design.py
+++ b/hashmap_design.py
@@ -1,3 +1,9 @@
+# Time Complexity : Hash map: O(1)
+# Space Complexity : O(n)
+# Did this code successfully run on Leetcode : it runs in IDE, but fails in leetcode
+# Any problem you faced while coding this : No
+
+
 class ListNode:
 
     def __init__(self, key, value, next=None):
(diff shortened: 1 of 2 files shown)


hashmap_design.py
```diff
@@ -0,0 +1,100 @@
+class ListNode:
```
Sumesh-H, line 7: Kindly follow the lecture. The logic is wrong.

min_stack.py
```diff
@@ -1,3 +1,9 @@
+# Time Complexity : Min Stack: O(1)
+# Space Complexity : O(n)
+# Did this code successfully run on Leetcode : it runs in IDE, but fails in leetcode
```
Sumesh-H, line 3: You should not specify size. 
def __init__(self):
        self.stack = []
        self.min_val = sys.maxsize
    def push(self, x: int) -> None:
        if x <= self.min_val:
            self.stack.append(self.min_val)
            self.min_val = x
            self.stack.append(x)

Sumesh-H:
Good work, Enna! Kindly check inline comments.

PR #245, opened 2020-08-16
Completed Design-1: Please review

asureban1:
Completed Design-1


Gupta-Keshav-001378461:
Good Job!!
For designing HashMap there is a better approach which you will learn in the class.
Explain your approach in layman terms.

PR #248, opened 2020-08-16
Completed Sprint Design-1

chinmayeetapaskar:


Gupta-Keshav-001378461:
Good Job!!
Explain the intuitive approach for the solution.

PR #262, opened 2020-08-17
Completed Big N Problem #1 - Please review. 

rimzimt:
Completed HashMap problem.


design1.py
```diff
@@ -0,0 +1,81 @@
...
+    def remove(self, key: int) -> None:
+        """
+        Removes the mapping of the specified value key if this map contains a mapping for the key
+        """
+        
+        hashValue=self.getHashValue(key)
+        
+        for index,element in enumerate(self.arr[hashValue]):
+            if element[0]==key:  #check key
+                self.arr[hashValue].pop(index)
```
samarthjj, line 71: Causes shifting to occur, O(n) operation.

design1.py
```diff
@@ -0,0 +1,81 @@
+# Time Complexity : O(N/k) k=fixed size of hashmap. N=no. of unique keys stored in hashmap
```
samarthjj, line 1: The worst case Time Complexity will be O(n^2). If all keys are stored in one single index and remove would cause shifting.

samarthjj:
Good job Rimzim, Please review the inline comments.

samarthjj:
Good job Rimzim, Please review the inline comments.

rimzimt:
Modified the problems with new approach discussed in class. Please review.

PR #265, opened 2020-08-17
Completed Design-1 problems - Please review.

vishalkumarmg1:


Himanshu-Agarwal93:
Do not use built-in data structures.
Please mention time and space complexities for each solution.

PR #278, opened 2020-08-18
Problem 1 - Implement HashMap Problem 2 - Find the Min Stack

desham05:


khushal1198:
Good intuitive HashMap Solution Sham.
Time complexity is correct but space complexity will be O(n) as we will be storing n number of elements in our HashMap.
Here you are creating the whole column with values -1 for adding a single value, which will be a constraint on space. Try implementing the solution from s30 dashboard. It's more optimized.

Good job with MinStack. Correct time complexity, but space complexity will be O(n) where n will be the number of elements in the stack.

PR #281, opened 2020-08-18
Design Minstack

rupeshbiradar40:
Design of minstack using single stack and a variable. this algorithm has space and time complexity of o(1) for all the operations including push, pop, min and top


Himanshu-Agarwal93:
Good job. Check the space complexity.

PR #297, opened 2020-08-30
Day1 Problems : Please review (TC, SC)

ShaivyT:


samarthjj:
Good job on both the problems!

samarthjj:
Good job on both the problems!

PR #304, opened 2020-09-14
Completed Spring Design 1- Please Review

AkshayKajale:


Lohith-Aswathappa (commented):
Good work, Akshay!
Try the solution taught in class as well.

Lohith-Aswathappa:
Good work, Akshay!
Try the solution taught in class as well.

PR #309, opened 2020-09-15
Completed Sprint Design-1 : Please Review

Agarka:


706.js
```diff
@@ -0,0 +1,102 @@
+// Time Complexity : 
+// Space Complexity :
```
kannan-mrugapphan, line 2: Ideally your run time is expected to be constant and its O(1) amortized. It gets better with more complex hash functions. Space is O(n) - you somehow have to store all elements

kannan-mrugapphan (commented):
Good Job

kannan-mrugapphan:
Good Job

PR #310, opened 2020-09-15
Design-1 Hashmap and Minstack

rajjiyani:


Problem1_Hashmap.java
```diff
@@ -0,0 +1,40 @@
+package design1;
+
```
kannan-mrugapphan, line 2: You will run out of memory when the number of elements becomes large. You should implement some HASH function and rehash later if load factor crosses threshold.

kannan-mrugapphan (commented):
Good Job with Min Stacks. Please re do the other problem

kannan-mrugapphan:
Good Job with Min Stacks. Please re do the other problem

PR #311, opened 2020-09-15
completed Design 1. A bit confused with the time complexity of minStack. Please review and comment.

yash071096:


designHashMap.cpp
```diff
@@ -0,0 +1,36 @@
+//time complexity:O(1)
+//space complexity:O(n)
+//executed on leetcode: yes
+//approach:using array
+//any issues faced? no
```
Lohith-Aswathappa, line 5: Try the approach taught in class.

Lohith-Aswathappa (commented):
Good work, Yogesh!
Check inline comment.

Lohith-Aswathappa:
Good work, Yogesh!
Check inline comment.

designHashMap.cpp
yash071096: Yes, i'll try it today.

PR #313, opened 2020-09-15
Design-1 Completed

yogesh696ksingh:


Problem1.py
```diff
@@ -0,0 +1,39 @@
+# Time Complexity :
+# Space Complexity : 
+# Did this code successfully run on Leetcode : Yes
+# Any problem you faced while coding this : Unsure about deriving time and space complexity
+
```
Lohith-Aswathappa, line 5: Include the complexities and try the approach taught in class.

Lohith-Aswathappa (commented):
Nice try, Yogesh!
Check inline comments.

Lohith-Aswathappa:
Nice try, Yogesh!
Check inline comments.

PR #314, opened 2020-09-15
Solutions for Design 1

sujaygijre:


DesignHashMap.cpp
```diff
@@ -0,0 +1,31 @@
+class MyHashMap {
+public:
+    /** Initialize your data structure here. */
```
Lohith-Aswathappa, line 3: Try the approach taught in class

Lohith-Aswathappa (commented):
Good work, Sujay!
Check the inline comments.

Lohith-Aswathappa:
Good work, Sujay!
Check the inline comments.

PR #321, opened 2020-09-16
"Finished Design-1 but with brute force Hashmap"

Jaisu-1:
Will try and improve on solution after today's lecture and attempt the other 2 problems as well.


Jaisu-1:
Hashmap was inspired from an existing solution as well.

PR #334, opened 2020-10-12
Completed Design-1- Please Review

khannakanika:


Problem_1.py
```diff
@@ -0,0 +1,75 @@
+"""
+Time Complexity : O(n)
+Space Complexity : O(n)
+Did this code successfully run on Leetcode : Yes
+Any problem you faced while coding this : I am having issue in designing a new hash function if I want to 
+apply double hashing
```
kannan-mrugapphan, line 6: Double hashing is essentially hash maps of hash maps, so the hash function for second map is also key % size of second map

kannan-mrugapphan (commented):
Good Job

kannan-mrugapphan:
Good Job

PR #337, opened 2020-10-13
Completed Day1-Design1 exercises - Please review.

bhumikatiwari16:


ramya-sruthi-d (commented):
Good job with your first submission. The whole idea of hashmap is to get elements in O(1) try to optimize it further.

ramya-sruthi-d:
Good job with your first submission. The whole idea of hashmap is to get elements in O(1) try to optimize it further.

bhumikatiwari16:
Reworked on both the problems as per discussion in today's class. Please review.

PR #343, opened 2020-10-13
Adding solutions for Design-1

UtsavRaychaudhuri:


design_hashmap.py
```diff
@@ -0,0 +1,69 @@
+# Time Complexity : O(n/c):- n=No. of elements, c=No. of slots avilable, 997
+# put:- average case:- O(1)/worst case:- O(n)
+# get:- average case:- O(1)/worst case:- O(n)
+# remove:- average case:- O(1)/worst case:- O(n)
+# Space Complexity :O(c):- c=No. of slots available
+# Did this code successfully run on Leetcode : Yes
+#  Any problem you faced while coding this : I wanted to implement Double Hashing, but couldn't figure out a 
+# way to find out what would be the best second hash function we should use for resolving collisions. Maybe a different prime number.
```
kannan-mrugapphan, line 8: Double hashing is essentially hash maps of hash maps under the hood. The second hash function is also key % size of inner map. The size of inner map can also be the same value. Just handle collisions accordingly.

design_minstack.py
```diff
@@ -0,0 +1,41 @@
+# Time Complexity :push:- O(1) pop:-O(1) top- O(1) getmin:- O(1)
+# Space Complexity : O(n)
+# Did this code successfully run on Leetcode : Yes
+# Any problem you faced while coding this : At first I thought of using a heap to keep track of the minimum element but that would give me O(logn) time
```
kannan-mrugapphan, line 4: Apart from that, the interviewer expects you to use only stacks here

kannan-mrugapphan (commented):
Good Job.

kannan-mrugapphan:
Good Job.

design_minstack.py
UtsavRaychaudhuri: @kannan-mrugapphan In python, I don't think we have a stack data structure. As far as I understood a stack doesn't use contiguous memory locations which an array which many people call stack in python does. The closest python comes to having a stack data structure is the deque in the collections framework which can be called a stack if don't use the popleft() method. Otherwise, I have to implement my own stack data structure. Please let me know your thoughts about this.

PR #344, opened 2020-10-13
Please help! I have doubts on both the problems

Prachal80:


Prachal80:
The doubts are solved now!

PR #345, opened 2020-10-13
Implemented Hash map and min stack 

pkariya1:
Please review


pkariya1:
Please review I have pushed with different github Id

PR #352, opened 2020-10-14
Created Desgn1_MinStack.java

jeevan3107:
Design1_MinStack problem


khushal1198 (commented):
Good work Jeevan. Please try the optimized approach where you only use one stack.

khushal1198:
Good work Jeevan. Please try the optimized approach where you only use one stack.

PR #353, opened 2020-10-14
Please review Design 1 Thanks!

PrajaktaKhandve:


khushal1198 (commented):
Good work Prajakta! Please try the more optimal approach for min stack where you use only one stack.

khushal1198:
Good work Prajakta! Please try the more optimal approach for min stack where you use only one stack.

PR #368, opened 2020-10-21
design1_submission

sujana99:


design1.py
```diff
@@ -0,0 +1,82 @@
...
+        self.minstack.append(self.minvalue)
+        
+
+    def pop(self) -> None:
+        self.minstack.pop()
+        self.minvalue=self.minstack[-1]
+        return self.instack.pop()
+
+    def top(self) -> int:
+        return self.instack[-1]
```
samarthjj, line 32: Please add edge cases, to prevent from pop() peek() on operating on an empty stack.

samarthjj:
Good work here Sujana, Please review the inline comments.
The HashMap implementation is not correct and not optimized, please rework your logic. You can text me on slack if you face any issues or need help.

samarthjj:
Good work here Sujana, Please review the inline comments.
The HashMap implementation is not correct and not optimized, please rework your logic. You can text me on slack if you face any issues or need help.

PR #370, opened 2020-10-21
Completed both problems

akshaydp1995:


Problem1.py.txt
```diff
@@ -0,0 +1,45 @@
+# Time Complexity : O(n^1/2) to initialize array??, O(1) put, O(1) get, O(1) remove
```
kannan-mrugapphan, line 1: Have it as null itself and initialize in put() if null is still there. This avoids that initialization issue.

kannan-mrugapphan (commented):
Good Job.

kannan-mrugapphan:
Good Job.

PR #374, opened 2020-11-16
Completed Design-1 please review

SubhashiniArun:


MinStack.java
```diff
@@ -0,0 +1,59 @@
...
+        stack.push(x);
+        if(minStack.isEmpty() || x<=minStack.peek()){
+            minStack.push(x);
+        }
+    }
+    
+    public int pop() {
+    	//if the top of the minstack equals the 
+        if(stack.peek() == minStack.peek()){
+           return minStack.pop();
```
samarthjj, line 30: Don't use return here, you would also have to pop from the main stack and this won't let the control flow reach the stack.pop() line.

samarthjj:
Good work here Subhashini, Please review the inline comments.

samarthjj:
Good work here Subhashini, Please review the inline comments.

PR #377, opened 2020-11-17
hashmap and min stack solved

rohank2002:


HashMap.java
```diff
@@ -0,0 +1,82 @@
+// Time Complexity :
+//Put O(n)
+//Remove O(n)
+//Get O(n)
```
rohank2002, line 4: O(1) for all.

PR #388, opened 2020-11-17
Added Problem 1 and Problem 2

devanshshah09:


Himanshu-Agarwal93 (commented):
Incorrect approach for hashmap

Himanshu-Agarwal93:
Incorrect approach for hashmap

PR #391, opened 2020-11-17
Design-1 Java implementation

ckallianpur:


MyHashMap.java
```diff
@@ -0,0 +1,101 @@
...
+        {
+            nodeArr[index].value = value;
+            return;
+        }
+        Node curr = nodeArr[index];
+        while(curr.next != null)
+        {
+            if (curr.next.key == key)
+            {
+                curr.next.value = value;
```
samarthjj, line 40: If value is updated, the function should return, otherwise line 42 will also execute creating an unwanted duplicate node.

samarthjj:
Good work Chaitra, Please review the inline comments.
Please add your Time & Space for all the problems, use template from sample.java file.
Please also add a brief description of your approach in 3-4 sentences for all the problems.

samarthjj:
Good work Chaitra, Please review the inline comments.
Please add your Time & Space for all the problems, use template from sample.java file.
Please also add a brief description of your approach in 3-4 sentences for all the problems.

PR #394, opened 2020-11-17
 Design stack problem Please review

souravit:


Sample.java
```diff
@@ -1,7 +1,51 @@
-// Time Complexity :
+// Time Complexity :0(1)
 // Space Complexity :
```
Lohith-Aswathappa, line 2: Include time and space complexities as per your understandings, we'll check afterwards

Lohith-Aswathappa (commented):
Nice try, Sourav!
Check inline comments

Lohith-Aswathappa:
Nice try, Sourav!
Check inline comments

PR #398, opened 2020-11-18
Add solution 1

pkulkarni01:


MyHashMap.java
```diff
@@ -0,0 +1,102 @@
+// Time Complexity : O(n)
+// Space Complexity : O(n)
+// Did this code successfully run on Leetcode : No
+// Any problem you faced while coding this :  Getting wrong answer
```
kannan-mrugapphan, line 4: Try to write a helper which gives position that has to be changed and call it appropriately. Please DM if still stuck after lecture.

kannan-mrugapphan (commented):
Good Job

kannan-mrugapphan:
Good Job

PR #407, opened 2020-12-17
Completed Sprint Design 1: Please review

shashwathkamath:
I think by mistake 2 Min_Stack assignment to pushed. One of them contains TC and SC mentioned. Please accept that.

### DIFF
--- a/MinStack.java
+++ b/MinStack.java
@@ -1,5 +1,8 @@
 class MinStack {
 
+    //Time Complexity: 1. O(1) to push the element 2. O(1) to pop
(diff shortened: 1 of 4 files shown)


samarthjj:
Good work Shashwath. Please also try implementing Min Stack problem using one stack.

samarthjj:
Good work Shashwath. Please also try implementing Min Stack problem using one stack.

kamathshashwath1991:
@samarthjj: I have implemented Min Stack using single stack. Please have a look

PR #410, opened 2020-12-21
Completed Design -1: Please review

juhi1296:


kannan-mrugapphan (commented):
Good Job. The first problem doesn't implement a hashing function to get the index. Please reach out if you are still confused after class. Please include the complexities as well.

kannan-mrugapphan:
Good Job. The first problem doesn't implement a hashing function to get the index. Please reach out if you are still confused after class. Please include the complexities as well.

juhi1296:
Hi @kannan-mrugapphan , thank you for your quick response. Yes, I need to understand that hashing part, I will try to implement a new code which includes hashing. I have mentioned the time and space complexity on the first and second line of code.

PR #416, opened 2020-12-21
Completed Design1 - Please Review

gangakalburgi:


MyHashMap.java
```diff
@@ -0,0 +1,113 @@
+// Time Complexity : O(1), for put, get and remove operations
+// Space Complexity : O(1000) or can say constant space O(1) since we are using list of size 1000
+// Did this code successfully run on Leetcode : yes
+// Any problem you faced while coding this : Have some doubts about time complexity
+
```
Lohith-Aswathappa, line 5: According to your logic, you are traversing through all the pairs in the bucket to put/update, get and remove. 
Since you cannot find the key at once, O(1) is not the correct time complexity. It becomes O(N) asymptotically. 

Also, since you have 1000 as a fixed size, it's not recommended as you cannot scale up. 

Try the solution as taught in the session.

Lohith-Aswathappa (commented):
Nice try, Ganga!
Review the inline comments and update the solutions

Lohith-Aswathappa:
Nice try, Ganga!
Review the inline comments and update the solution

MyHashMap.java
gangakalburgi: Yes, I agree. Will do the changes. Thank you.

PR #422, opened 2020-12-22
Please review Design-1

Lohita1114:


DesignMinStack.java
```diff
@@ -0,0 +1,59 @@
+/* package codechef; // don't place package name! */
+
+import java.util.*;
+import java.lang.*;
+import java.io.*;
+
+class MinStack {
```
samarthjj, line 7: Please error guard pop(), top() & getMin() methods to prevent from operating on an empty stack, the program will crash otherwise.

samarthjj:
Good work Lohita, Please review the inline comments.
Please also try implementing HashMap problem in a more controlled manner like done in class.

samarthjj:
Good work Lohita, Please review the inline comments.
Please also try implementing HashMap problem in a more controlled manner like done in class.

PR #424, opened 2020-12-22
Design 1 - Please review

snehasg95:
For the first problem - hash Map 
I attempted the hashing concept since my first implementation had a large space complexity, but have doubts about the removal method as well as calculating time-space complexities for this approach

### DIFF
--- a/hashmap_design.py
+++ b/hashmap_design.py
@@ -1,0 +1,123 @@
+# Approach: Initially tried with using a large array defined as per constraint ([0, 1000000])
+# If the key is not found we return -1 hence intialized the huge sized array with -1
+# Time complexity - O(1) for get, put,remove
+# Space complexity - O(N) where N is the size of the array
+# Not an ideal solution as defining a huge array adds to space constraint 
+# Ran with 365 ms and 39 Mb space
+class MyHashMap:
+
+    def __init__(self):
+        """
+        Initialize your data structure here.
+        """
+        self.arr = [-1] * 1000000
+        
+
+    def put(self, key: int, value: int) -> None:
+        """
+        value will always be non-negative.
+        """
+        self.arr[key] = value
+        
+
+    def get(self, key: int) -> int:
+        """
+        Returns the value to which the specified key is mapped, or -1 if this map contains no mapping for the key
+        """
+        
+        return self.arr[key]
+        
+
+    def remove(self, key: int) -> None:
+        """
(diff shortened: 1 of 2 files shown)


min_stack.py
```diff
@@ -0,0 +1,73 @@
...
+# Time Complexity : O(1) # append, pop peek all are constant time operations
+# Space Complexity : O(1) ## append, pop peek all are constant space operations
+# Succesfully ran in leetcode
+# Base Case: Not considering an empty stack as constraint mentions non-empty stack
+
+# Doubt: I have declared a minStack [] but that only holds the min value appended for every push method. 
+# I mentioned O(1), since if methods push/pop are run independantly it is hardly storing one value(append) or popping one val at a time, which is constant space/time
+# Should I mention this as the case or the overall when we need to perform on 'n' such elements in the stack? Not quite sure
+
+class MinStack:
```
samarthjj, line 14: Please error guard pop(), top() & getMin() methods to prevent from operating on an empty stack, the program will crash otherwise.

hashmap_design.py
```diff
@@ -0,0 +1,123 @@
...
+# obj = MyHashMap()
+# obj.put(key,value)
+# param_2 = obj.get(key)
+# obj.remove(key)
+
+#####
+
+# Approach 2: Declared an array for holding the data with the help of a linked list
+# Used a basic hashing by performing modulo operation on the key wrt to the size declared with a linked list approach
+# Not sure how to calculate Time complexity
```
samarthjj, line 49: The worst case Time complexity will be linear. If assumed that all buckets hold equal number of elements, the Asymptotic Time here will be O(n/m), otherwise in the worst case if all elements are in a single bucket then O(n).

samarthjj:
Good work here Sneha, Please review the inline comments.
Please also try implementing the Min Stack problem using one stack.

samarthjj:
Good work here Sneha, Please review the inline comments.
Please also try implementing the Min Stack problem using one stack.

min_stack.py
snehasg95: Thanks for the feedback, i referred to the constraint on leetcode and it said `Methods pop, top and getMin operations will always be called on non-empty stacks.` 

In this case should I still address the constraint?

samarthjj: Yes, it is always good to add on edge cases that Leetcode might or might not suggest, ultimately you will not be coding on a platform like leetcode for your interviews. For any further doubts please message me on Slack.

PR #427, opened 2020-12-22
Design-1 Lokesh

LokeshVeluswamy:


MinStack.java
```diff
@@ -0,0 +1,39 @@
+// Time Complexity : O(1)
+// Space Complexity : O(1)
```
samarthjj, line 2: Space: O(n)

MinStack.java
```diff
@@ -0,0 +1,39 @@
...
+// Did this code successfully run on Leetcode : Yes
+// Any problem you faced while coding this : No
+
+class MinStack {
```
samarthjj, line 6: Please error guard pop(), top() & getMin() methods against operating on an empty stack, the program will crash otherwise in certain test cases.

samarthjj:
Good work here Lokesh, Please review the inline comments.
Please also try implementing the Min Stack problem using one stack.

samarthjj:
Good work here Lokesh, Please review the inline comments.
Please also try implementing the Min Stack problem using one stack.

PR #443, opened 2020-12-25
Design 1 please review 

Saphal-Thammu:
Yes I have few concerns. Firstly I want to know the code which worked in Leet code is giving me errors
in my system. The errors are :
1) error: a space is required between consecutive right angle brackets (use '> >')
    vector<vector<pair<int,int>>> m;
2)error: expected expression
        m.resize(1000,{});
3)error: expected expression
        m[hashnum].push_back({key,value});

        Also I would like to know how should be the main function be the test.cpp has main function,
        which I did not know how to link and execute

Please can we set up a quick call whenever you are free so that my doubts be clarified.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,5 @@
+{
+    "files.associations": {
+        "vector": "cpp"
+    }
+}
--- a/Design1_Hashmap.cpp
+++ b/Design1_Hashmap.cpp
@@ -1,0 +1,85 @@
+// Time Complexity : O(1)
+// Space Complexity : O(1000)
+// Did this code successfully run on Leetcode : yes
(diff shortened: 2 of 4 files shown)


super30admin:
We resolved this yesterday ..right ?

`m.resize(1000,{});` is not the expression to be used for vector. {} is array in c++ and not a vector

super30admin:
```error: a space is required between consecutive right angle brackets (use '> >')```

This error is self explanatory. Please put a space between > brackets

Saphal-Thammu:
yeah but I wanted to know why it did not give me an error in Leet code but when I tried to run it in my system on terminal. It gave me errors.

super30admin:
Ping me on slack ...Jaspinder

PR #445, opened 2020-12-26
Completed Design-1 -please review

fnoor2:


maithrreye:
Good Job ! I don't see problem with Hashmap solution however it didn't work. Can you try using a linklist instead of array and see if there is no memory issue .

maithrreye:
Good Job ! I don't see problem with Hashmap solution however it didn't work. Can you try using a linklist instead of array and see if there is no memory issue .

PR #469, opened 2021-01-20
Completed Design 1 - Please Review

srirachanaachyuthuni:


kannan-mrugapphan (commented):
Try to implement a hashing function to get the key(index) and store the value accordingly. For the second problem, try to use a support stack to save on time complexity.

kannan-mrugapphan:
Try to implement a hashing function to get the key(index) and store the value accordingly. For the second problem, try to use a support stack to save on time complexity.

PR #471, opened 2021-01-21
design 1 please review

jammulanikhil1:


kannan-mrugapphan (commented):
Please use a hashing method to get the key (index) to store the respective value. Please submit the other problem too.

kannan-mrugapphan:
Please use a hashing method to get the key (index) to store the respective value. Please submit the other problem too.

PR #485, opened 2021-02-16
Completed Design-1, Please Review

kaushikguntaka:


kannan-mrugapphan (commented):
Good Job. For the first problem, implement a hashing function that gives an appropriate index. Do not hard code array size at the first.

kannan-mrugapphan:
Good Job. For the first problem, implement a hashing function that gives an appropriate index. Do not hard code array size at the first.

PR #486, opened 2021-02-16
Completed Design 1 - Please review

Mani632:


kannan-mrugapphan (commented):
For the hashmap problem, use a hashing function to get the index. Do not hardcode sizes

kannan-mrugapphan:
For the hashmap problem, use a hashing function to get the index. Do not hardcode sizes

PR #494, opened 2021-02-16
Completed min stack - Please review

Manudeep-git:
I couldn't comprehend Hashmap design, I will go through it once again after the class today and update this PR


Manudeep-git:
Completed Hashmap, please review, I was bit confused about TC and SC, Please let me know if my understanding is correct

Manudeep-git:
> I couldn't comprehend Hashmap design, I will go through it once again after the class today and update this PR

02/18
Completed both problems, please review

PR #503, opened 2021-02-17
Design 1, Problem 2:Min Stack . Please review

sachin0112358:


MinStack.java
```diff
@@ -0,0 +1,50 @@
+class MinStack {
+
+    /** initialize your data structure here. */
+    Stack<Integer> s;
```
yahska-ednihs, line 4: Good Job! Please add comments, time and space complexity. Also please add the other problem for this assignment.

PR #524, opened 2021-03-16
Completed Sprint Design 1 - Please review

vp43:


Problem1.py
```diff
@@ -0,0 +1,101 @@
...
+
+    def __init__(self):
+        """
+        Initialize your data structure here.
+        """
+        #hashmap of fixed size - array
+        self.hashmap = [None]*1000
+        
+    def _hash(self,data):
+        return data%10
```
sibalnirbhay, line 27: You have a bucket of size 1000 while you're actually using only the first 10 blocks of it. Why is that?

sibalnirbhay (commented):
Looks good! Read inline comments about your HashMap implementation.

sibalnirbhay:
Looks good! Read inline comments about your HashMap implementation.

PR #551, opened 2021-04-20
Completed Design1 - Please review

Sahulphan:


maithrreye:
Good Job Sahul! Please add language file instead of txt file.

maithrreye:
Good Job Sahul! Please add language file instead of txt file.

PR #557, opened 2021-04-21
Min_Stack is Completed Please Review

virendra26893:
Time  Complexity: o(1)
Space  Complexity: o(n)

Description:
Created two stacks:
1. base stack: this will store all the values.
2.  minStatus: this will hold the minimum value corresponding to each value in the base stack which is a one to one mapping.
3. then created a min variable which will hold the minimum values at each push and pop operation.

Logic of:

1. push() : push all the values in the base stack without any check but while pushing the same values we need to make sure the if the current value is the less than the current min then only we need to push that value in the minStatus stack and update the min also with that value else we have update the with previous min value.
2. pop(): remove and return the top element from both base and minStatus stack and update the min with the peek of the minStatus element as that will be logically next min among remaining elements. 
Note: if the minStatus stack become empty then update min with Integer.MAX_VALUE
3. getmin(): return min value directly.
4. gettop: return the top of base stack.
5.  initially assign min value with Integer.MAX_VALUE.

### DIFF
--- a/solutions/155. Min Stack.java
+++ b/solutions/155. Min Stack.java
@@ -1,0 +1,50 @@
+class MinStack {
+    
+    private Stack<Integer> base;
+    private Stack<Integer> minStatus;
+    int min;
+​
+    /** initialize your data structure here. */
+    public MinStack() {
+        base=new Stack<Integer>();
+        minStatus=new Stack<Integer>();
+        min=Integer.MAX_VALUE;
+    }
+    
+    public void push(int val) {
+        base.push(val);
+        if(min>val){
+            min=val;
+            minStatus.push(val);
+        }else{
+            minStatus.push(min);
+        }
+    }
+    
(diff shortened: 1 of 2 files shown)


solutions/155. Min Stack.java
```diff
@@ -0,0 +1,50 @@
...
+    
+    public int top() {
+       return base.peek();
+    }
+    
+    public int getMin() {
+        return min;
+    }
+}
+​
```
Lohith-Aswathappa, line 42: include all the comments and complexities inside the code.
Don't comment on the commit. Creates an issue with PR status internally.

Lohith-Aswathappa (commented):
Good work, Virendra!
Check inline comment and try other problem as well.

Lohith-Aswathappa:
Good work, Virendra!
Check inline comment and try other problem as well.

PR #585, opened 2021-05-24
Complete Sprint Design-1 : Please review

chaitanya003:


hash_map.cpp
```diff
@@ -0,0 +1,40 @@
+// Time Complexity : O(1)
+// Space Complexity : O(n)
```
sridhar-anish, line 2: If You are using constant space Space Complexity would be O(1)

hash_map.cpp
```diff
@@ -0,0 +1,40 @@
+// Time Complexity : O(1)
+// Space Complexity : O(n) 
+// Did this code successfully run on Leetcode : Yes
```
sridhar-anish, line 3: If you are using Constant space, space complexity would be O(1)

sridhar-anish (commented):
Nice try Chaitanya, make sure to try out the solutions taught in class as well

sridhar-anish:
Nice try Chaitanya, make sure to try out the solutions taught in class as well

PR #624, opened 2021-05-26
Completed Design1

Ishu-jha:


maithrreye:
Good Job Ishu ! Please do the other problem as well. Please add the problems as seperate language file !

maithrreye:
Good Job Ishu ! Please do the other problem as well. Please add the problems as seperate language file !

PR #647, opened 2021-05-27
completed Design-1 Probs. Please Review

devarshamin98:


maithrreye:
Good job on Minstack problem ! Please add the time and space complexity and comment about the approach of the problems. In the Hashmap code , temp=bucket(k), May I know where the bucket is defined? are u using anysort of package? this was little confusing . A better structure of code with comments would do good.

maithrreye:
Good job on Minstack problem ! Please add the time and space complexity and comment about the approach of the problems. In the Hashmap code , temp=bucket(k), May I know where the bucket is defined? are u using anysort of package? this was little confusing . A better structure of code with comments would do good.

PR #660, opened 2021-06-01
Hashmap + MinStack

hharshitarora:
So for HashMap I pass 34/36 tests - The case where 11,0 was added, since array default value is 0 that's why it fails. The idea I had was to initialize all values to -1. Please let me know if this works? 
For MinStack i used the double stack method, i tried the single stack method but could not succeed.

### DIFF
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -1,0 +1,8 @@
+# Default ignored files
+/shelf/
+/workspace.xml
+# Datasource local storage ignored files
+/dataSources/
+/dataSources.local.xml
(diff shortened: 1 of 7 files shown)


khushal1198 (commented):
Didn't understand the doubt in the HashMap. Since the question is covered in class now, hope the doubt is cleared. If not, please reach out your TA.

khushal1198:
Didn't understand the doubt in the HashMap. Since the question is covered in class now, hope the doubt is cleared. If not, please reach out your TA.

PR #675, opened 2021-06-22
Completed Sprint Design - please review

Mangesh-Ashtankar:


problem1.py
```diff
@@ -0,0 +1,65 @@
+class MyHashMap:
+    
+    def __init__(self):
+        """
+        Initialize your data structure here.
+        """
+        self.array = [-1 for i in range(1000000)]
+
```
Lohith-Aswathappa, line 8: Nice try, Mangesh! But try the approach taught in class as well. Helps you better understand

Lohith-Aswathappa (commented):
Good work, Mangesh!
check inline comment

Lohith-Aswathappa:
Good work, Mangesh!
check inline comment

PR #677, opened 2021-06-22
completed brute force

venkatagudimela:
I have completed brute force for both problems, please verify


ramya-sruthi-d (commented):
Good job with code, please mention the complexities as well.

ramya-sruthi-d:
Good job with code, please mention the complexities as well.

PR #693, opened 2021-06-23
HashMap design done using brute's force

SurbhiZambad2510:


Rooppesh:
Good Job Surbhi! Please add the TC and SC

Rooppesh:
Good Job Surbhi! Please add the TC and SC

PR #710, opened 2021-07-20
completed in python in two new files hashmap.py and minstack.py

psambalkar:


Rooppesh:
Good Job! Everything looks good!

Rooppesh:
Good Job! Everything looks good!

PR #747, opened 2021-08-17
Design-1 complete please review

rajchoksi1997:


maithrreye:
Good Job Raj ! 
Your understanding about space complexity written in Min stack code is correct . You have a stack , you put N numbers inside the stack and return output just the Minimum value. So stack is auxiliary space. you report them in space complexity.

maithrreye:
Good Job Raj ! 
Your understanding about space complexity written in Min stack code is correct . You have a stack , you put N numbers inside the stack and return output just the Minimum value. So stack is auxiliary space. you report them in space complexity.

PR #796, opened 2021-08-20
Create hasmap.py

MaheshReddy108:


sridhar-anish (commented):
Good Job with everything, make sure to write TC and SC, try out the other question as well

sridhar-anish:
Good Job with everything, make sure to write TC and SC, try out the other question as well

MaheshReddy108:
sure, Thanks @sridhar-anish , TC mean Time complexity ? what does SC mean ?

PR #805, opened 2021-08-24
Both the problems

ami-doshi:
I tried Hashmap with Double Hashing instead of Chaining, I keep getting the wrong answer, unable to debug where I am going wrong.


ami-doshi:
Original Solution for HashMap provided wrong solution. I wrote a new code. Please confirm if the complexities are correct

PR #817, opened 2021-09-14
Completed Design1 - Problem 2

ssvel04:


Rooppesh:
Good Job! Try the other problem as well!

Rooppesh:
Good Job! Try the other problem as well!

PR #818, opened 2021-09-14
Completed Min-Stack- Please Review

7shradhapawar:


Pinak08 (commented):
No problem you can take help if you don't get it.
Practice more problems to get logic more frequently.
I think space complexity can be said as O(1)
As are building stack is a part of problem but in the approach we do not use any extra space.

Pinak08:
No problem you can take help if you don't get it.
Practice more problems to get logic more frequently.
I think space complexity can be said as O(1)
As are building stack is a part of problem but in the approach we do not use any extra space.

PR #831, opened 2021-09-14
Completed MinStack 

ManjariK0308:
MinStack => Problem of Design 1


brunda4:
Good work on Stack. Please update the hashmap code such that you dont use hashmap to design a hashmap

PR #865, opened 2021-10-07
Completed design1 - please review

ikshaku:
Both problems solved on leetcode successfully.


brunda4 (commented):
Good Work on linked list. You have used stack to design a stack, which is not correct approach.

PR #945, opened 2021-12-22
Completed Design-1 MinStack problem - Please Review

swaroopguptaba:
Implemented Min stack with push, pop, isEmpty, isFull() and getMin() operations.
Please let me know if anything can be improved

### DIFF
--- a/MinStack.java
+++ b/MinStack.java
@@ -1,0 +1,94 @@
+// Time Complexity : O(1)
+// Space Complexity : O(n)
+// Did this code successfully run on Leetcode: Yes
+// Any problem you faced while coding this: Took and initial approach of maintaining only 2 minimums
+// and that failed when there are 2 min elements with same value
+
+// Your code here along with comments explaining your approach
+
+import java.util.Stack;
+
+public class MinStack {
+
+    Stack<Integer> stack;
+    Stack<Integer> mStack; // to keep track of minimum's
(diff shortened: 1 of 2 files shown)


sonalgupta276 (commented):
Good job! Update after discussion in class. Mention TC and SC for each function

sonalgupta276:
Good job! Update after discussion in class. Mention TC and SC for each function

swaroopguptaba:
> Good job! Update after discussion in class. Mention TC and SC for each function

Hi Sonal,

Could you please let me know what do you mean by Update after discussion in class?

Should I be updating the above code with what's covered in class?

sonalgupta276:
Yes, you can after it's done in class if your approach is different. I think hashmap problem was not there when I last viewed

swaroopguptaba:
> Yes, you can after it's done in class if your approach is different. I think hashmap problem was not there when I last viewed

Got it, thank you.

Yes, Hashmap was not there last time. In these scenarios, where I will be pushing code to same branch, how should I proceed to submit the PR?

sonalgupta276:
Push commit to the same branch. It won't be viewed again most probably, but you can have code for your own sake

swaroopguptaba:
> Push commit to the same branch. It won't be viewed again most probably, but you can have code for your own sake

Okay, thanks

PR #974, opened 2021-12-27
completed minstack - please review

hiteshmadapathi:


sonalgupta276 (commented):
Follow as discussed in class

sonalgupta276:
Follow as discussed in class. Make different files for each problem

PR #990, opened 2022-01-31
Adding MinStack solution

anuragis2:


MinStack.java
```diff
@@ -0,0 +1,49 @@
+// Time Complexity O(1)
+// Space Complexity O(1)
+
```
shrivatsav, line 3: Please check the Space Complexity. We are storing `min`, which in worse case can be O(n) where n is the number of items in the stack.

shrivatsav (commented):
Good Work!

shrivatsav:
Good Work!

PR #999, opened 2022-02-01
Solved MinStack - Please Review

princerokn:


princerokn:
Solved Design HashMap - Please Review

brunda4 (commented):
Please provide complete solution for hash.

brunda4:
Please provide complete solution for hash.

princerokn:
> Please provide complete solution for hash.

Hi, Its a complete solution already. I am doing hashing using single array, where index is key and value is stored in the array. Its works fine. Do you want me to solve it with different approach?

PR #1010, opened 2022-02-03
completed design 1, please review

shiva-pa:
Please review


design_hashset.py
```diff
@@ -0,0 +1,66 @@
+## // Time Complexity :average time complexity O(1) and in worst case O(C) length of chain 
+# // Space Complexity :O(n)
```
shrivatsav, line 2: Good! Please try to implement double hashing and choose the `self.totalbucket` value in a suitable way where is we can say access time is O(1).

shrivatsav (commented):
Good Work!

shrivatsav:
Good Work!

PR #1034, opened 2022-03-01
Completed Design-1 : Please review

krinapatel:


minStack.js
```diff
@@ -0,0 +1,54 @@
+// Time Complexity : Time complexity is provided after each functions
+
+// Did this code successfully run on Leetcode : Yes code ran successfully on Leetcode
+
+// Any problem you faced while coding this : Not an issue exactly but I couldn't figure out the way of having O(1) 
+// time complexity for getMin function
+
+var MinStack = function() {
```
shrivatsav, line 8: You can use 2 lists, one to store the elements and other to store the corresponding min element. You can also use a single stack to handle this. So time complexity will be O(1) for `getMin()` but the space will be O(2n) or O(n) . 

Please try this approach.

shrivatsav (commented):
Good Work! Please address the feedback.

PR #1038, opened 2022-03-01
MinStack implemented - please review

soumya-94:


MinStack.java
```diff
@@ -0,0 +1,53 @@
+// Time Complexity : O(1) for each function
+// Space Complexity : O(n) where n is the number of values in the stack
+// Did this code successfully run on Leetcode : yes
+// Any problem you faced while coding this : no
+
+public class MinStack
```
shrivatsav, line 6: Please describe your approach briefly.

shrivatsav (commented):
Good Work!

shrivatsav:
Good Work!

MinStack.java
soumya-94: I have created a custom data structure called Node to contain the values inserted in stack. Each node has the value, min value till that element and a pointer to the next node in stack. In the push method, we keep adding nodes to the front of the head.

PR #1040, opened 2022-03-01
MinStack done, please review

nisargushah:


keshav-Gupta-IS (commented):
I would suggest either using 2 stacks or keeping track of minimum while pushing the element that way it will be O(1).

keshav-Gupta-IS:
I would suggest either using 2 stacks or keeping track of minimum while pushing the element that way it will be O(1).

PR #1091, opened 2022-03-30
completed problems 1 & 2

smruthyunjaya:
Please review


brunda4 (commented):
Please provide optimal solution for both problems.

brunda4:
Please provide optimal solution for both problems.

PR #1093, opened 2022-03-30
I have updated the sample file with code

akshatsrivastav53:
This is for design1


sanjayvarma199 (commented):
Good Job!! Please submit other problems as well

sanjayvarma199:
Good Job!! Please submit other problems as well

PR #1126, opened 2022-04-27
Completed Design-1 problems Please review

evolutionharsh:


keshav-Gupta-IS (commented):
Your remove function will not be O(1) also this implementation is will be okay for brute force. I would recommend submit the solution discussed in lecture.

keshav-Gupta-IS:
Your remove function will not be O(1) also this implementation is will be okay for brute force. I would recommend submit the solution discussed in lecture.

PR #1131, opened 2022-04-27
Completed S30 FAANMG Problem #2 {Medium} MinStack Design. Please Review

SaurabhGujare:
Design a SplStack that supports all the stack operations push, pop, isEmpty, isFull and an operation getMin() which should return minimum element from the SplStack. All these operations must be O(1). To implement this, you should only use standard Stack data structure.


meghasn (commented):
Good Job! Cross check after the discussion in class

meghasn:
Good Job! Cross check after the discussion in class

PR #1170, opened 2022-05-24
Completed Design-1

praneeha:


niharikagupta2021 (commented):
Good Job! Please add design-1 and design -2 problems separately or else PR status won't get updated

niharikagupta2021:
Good Job! Please add design-1 and design -2 problems separately or else PR status won't get updated

PR #1215, opened 2022-05-26
design 1 completed pls review

Aksh-Narayana:


keshav-Gupta-IS (commented):
The point of trying these problems before class is to develop the thought process. Refrain from looking at solutions online, the solution will be discussed in class.

keshav-Gupta-IS:
The point of trying these problems before class is to develop the thought process. Refrain from looking at solutions online, the solution will be discussed in class.

PR #1258, opened 2022-06-22
Completed Design1

shreyachaudhari99:


niharikagupta2021 (commented):
Good Job! Please add sc and tc for minstack as well

niharikagupta2021:
Good Job! Please add sc and tc for minstack as well

PR #1271, opened 2022-06-22
Design 01

soni-dhruv:


soni-dhruv:
I've committed my code in wrong way
But I'll upload it in correct way from now on

Thank you

Sumesh-H (commented):
This is good work. Please include TC & SC along with 3 point description.

Sumesh-H:
This is good work. Please include TC & SC along with 3 point description.

PR #1323, opened 2022-07-19
Update Sample.java

Sreelekhasamanu:
// Time Complexity : O(1) for each operation 
// Space Complexity : O(n) arraylist is used
// Did this code successfully run on Leetcode : yes
// Any problem you faced while coding this : 403 error while pushing it. so I added directly 
// Your code here along with comments explaining your approach
// Design a HashSet without using any built-in hash table libraries.
// Implement MyHashSet class:
// void add(key) Inserts the value key into the HashSet.
// bool contains(key) Returns whether the value key exists in the HashSet or not.
// void remove(key) Removes the value key in the HashSet. If key does not exist in the HashSet, do nothing.

import java.util.*; 
class hash{
        static class MyHashSet {
        static ArrayList<Integer> arr;
        public MyHashSet() {
            arr=new ArrayList<>();
        }

        public void add(int key) {
            if(!arr.contains(key))
                arr.add(key);

        }

        public void remove(int key) {
             if(arr.contains(key))
                 arr.remove(arr.indexOf(key));
        }

            return arr.contains(key);
        }
    public static void main(String [] args){
        MyHashSet hashset = new MyHashSet();
    hashset.add(1);         
    hashset.add(2);         
    hashset.contains(1);    // returns true
    hashset.contains(3);    // returns false (not found)
    hashset.add(2);          
    hashset.contains(2);    // returns true
    hashset.remove(2);          
    hashset.contains(2);    // returns false (already removed)
    }
}
}

### DIFF
--- a/Sample.java
+++ b/Sample.java
@@ -1,7 +1,62 @@
-// Time Complexity :
-// Space Complexity :
-// Did this code successfully run on Leetcode :
-// Any problem you faced while coding this :
 
+import java.util.Stack;
 
-// Your code here along with comments explaining your approach
+class MinStack {
+    
+     // Time Complexity:: push() - O(1), pop - O(1), top() - O(1), getMin() - O(1)
+     // Space Complexity:: O(n) 
+     // The code ran successfully on leetcode
+
+    Stack<Node> stack;
+
+    // Use Node class to push objects in the stack
+    public MinStack() {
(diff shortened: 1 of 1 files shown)


daredevilteja:
Please create new files for the problems and then raise the PR again. Don't send the code in comments.

PR #1332, opened 2022-07-20
Design 1 - please review it

vishnuvryeruva:


ishika2741:
Good work! Kindly include tc and sc. Submit other problem as well.

PR #1421, opened 2022-09-07
Design-1 Completed

samritabakshi:


Problem-1.js
```diff
@@ -0,0 +1,19 @@
+//Constant Time for all
```
jeelpatel18, line 1: You have to implement HashMap without use of in-built hash map

jeelpatel18 (commented):
Good work. Review the comments too.

jeelpatel18:
Good work. Review the comments too.

PR #1479, opened 2022-10-05
Completed Sprint Design-1 - Please Review

srin836:


problem_1.cpp
```diff
@@ -0,0 +1,51 @@
...
+// Any problem you faced while coding this : initialising 2D vector in c++ along with constructor. Jaspinder cleared the doubts.
+
+class MyHashSet {
+    int num_buckets;
+    int num_bucket_items;
+    vector<vector<bool>> st;
+    
+public:
+    MyHashSet() {  
+        num_buckets = 1000;
```
daredevilteja, line 13: change this to `num_buckets = 1001`, then it will work. You are getting max buffer overflow because you giving an input on 1000000 which is not possible with your defined sizes, you have to increase your `num_buckets` to 1001 to accommodate this

daredevilteja (changes requested):
Please do the requested changes

daredevilteja:
Good Work! Keep it up

PR #1572, opened 2023-01-04
Completed Design-1

shastry9940:


Problem 2
```diff
@@ -0,0 +1,45 @@
+// Explanation - Use 2 Stacks - to compare the current and the items that are puched into the stack
+// Time complexity = O(1)
+// Space complexity = O(1)
```
daredevilteja, line 3: SC should be O(n) because you are using extra stacks

daredevilteja (changes requested):

daredevilteja:
Good Work! Keep it up.

PR #1721, opened 2023-06-28
Completed Design-1

Tanushree1403:


ishika2741 (commented):
Well done! Kindly add TC and SC for both the problems as well.

ishika2741:
Well done! Kindly add TC and SC for both the problems as well.

PR #1768, opened 2023-08-04
Create PR for Design-1 

RudranilMitra:


Nandha2Tech (commented):
Good Job! A small suggestion - Since we are maintaining a minValue, itcan be used to return in getMin instead of minStack.peek();

Nandha2Tech (commented):
Good Job! Since nodes are created space complexity will be O(n).

Nandha2Tech:
Good Job! A small suggestion - Since we are maintaining a minValue, itcan be used to return in getMin instead of minStack.peek();
Good Job! Since nodes are created space complexity will be O(n).

PR #1790, opened 2023-08-30
Completed Design1

harichandanamone:
I have completed the Design1 but I have doubt when it comes to DesignHashSet as some of the test cases were failing.
Please provide insights on the code as its was failing for 5 test cases.

### DIFF
--- a/MinStack.java
+++ b/MinStack.java
@@ -1,0 +1,49 @@
+//Time complexity : all push,pop,top and getMin take O(1) Tc
+//Space Complexity : O(N) as we maintain two stacks
+// All test cases passed in Leetcode
+import java.util.Stack;
+
+class MinStack {
+    Stack<Integer> stack;
+    Stack<Integer> minStack;
+    int minValue;
+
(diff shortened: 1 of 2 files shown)


Nandha2Tech:
While adding, we have to check whether bucket is already there. And also for key=1000000 which is the upper limit specified in the problem, we will be needing one extra bucket. 
If key = 1000000, 1000000/1000 implies go to the 0th bucket and 1000000%1000 will be 1000 which will access 1000th bucketItem which is not there. So for 0th bucket we have bucketArray size as bucketArray+1.
Below are the changes that needs to be done. It will pass all test cases.
 public void add(int key) {
        int bucket = getBucket(key);
        int bucketItem = getBucketItem(key);
        if(array[bucket] == null)
        {
                  if(bucket == 0){
                      array[bucket] = new boolean[bucketItems+1];
                  }
                  else{
                      array[bucket] = new boolean[bucketItems];
                  }
        }
        array[bucket][bucketItem] = true;
    }

PR #1791, opened 2023-08-30
Completed Design-1 (HashSet and MinStack) - Please review

jailashah:
The solution to design minStack has space complexity of O(N). Is there a better approach available than this?


Nandha2Tech:
Great Work!
In design minStack, for push, pop, top operations we need a data structure to store the values. So SC of O(N) will occur.

Nandha2Tech:
Great Work!
In design minStack, for push, pop, top operations we need a data structure to store the values. So SC of O(N) will occur.

PR #1826, opened 2023-09-27
Design- 1 Completed

harshpatel226:


Nandha2Tech (commented):
Great Work! SC for both the solutions would be O(n).

Nandha2Tech:
Great Work! SC for both the solutions would be O(n).

PR #1830, opened 2023-09-28
Create Implementing_HashSet.py

KrishanuSinha:
Design a HashSet without using any built-in hash table libraries.

Implement MyHashSet class:

void add(key) Inserts the value key into the HashSet. bool contains(key) Returns whether the value key exists in the HashSet or not. void remove(key) Removes the value key in the HashSet. If key does not exist in the HashSet, do nothing.

### DIFF
--- a/Implementing_HashSet.py
+++ b/Implementing_HashSet.py
@@ -1,0 +1,36 @@
+class ListNode:
+    def __init__(self, key):
+        self.key = key
+        self.next = None
+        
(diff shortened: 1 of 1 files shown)


keshav-Gupta-IS (commented):
Good Job, Krishanu!! Keep it up.

keshav-Gupta-IS:
Good Job, Krishanu!! Keep it up.

PR #1854, opened 2023-11-08
Design-1 completed

Sharvika06:


Nandha2Tech (commented):
Great Work! For the problem design hashset, try to implement it without using any built in hashing data structure. Set in python is a built in data structure which uses hashing under the hood.

Nandha2Tech:
Great Work! For the problem design hashset, try to implement it without using any built in hashing data structure. Set in python is a built in data structure which uses hashing under the hood.

PR #1864, opened 2023-11-09
Completed Problem 2 and 3

BhagAneesh:
Sorry, I have created a PR combining two problem solutions. Not sure this is the correct way.


ishika2741 (commented):
Good work!  Kindly ass TC and SC as well.

ishika2741:
Good work!  Kindly ass TC and SC as well.

PR #1865, opened 2023-11-09
MinStack.java

aditird:
MinStack

### DIFF
--- a/MinStack.java
+++ b/MinStack.java
@@ -1,0 +1,45 @@
+// Time Complexity :
+// Space Complexity :
+// Did this code successfully run on Leetcode :
+// Any problem you faced while coding this :
+
+
+// Your code here along with comments explaining your approach
+class MinStack {
+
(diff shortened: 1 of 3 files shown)


Nandha2Tech (commented):
Great Work! Try to include TC and SC.
In MinStack Problem, priority queue is used. In priority queue, for each and every push and pop elements will be ordered based on the priority. So TC would become O(log k). Don't use pq to get the minimum. Have a min variable and update it.

In HashSet problem, handle the edge case when a key of value 1000000 arrives.

Nandha2Tech:
Great Work! Try to include TC and SC.
In MinStack Problem, priority queue is used. In priority queue, for each and every push and pop elements will be ordered based on the priority. So TC would become O(log k). Don't use pq to get the minimum. Have a min variable and update it.

In HashSet problem, handle the edge case when a key of value 1000000 arrives.

PR #1875, opened 2023-12-20
Design-1 Complete

Vd1299:


Hashmap design.py
```diff
@@ -0,0 +1,43 @@
+# Time Complexity : O(n)
```
ishika2741, line 1: You are taking bucket size of 1000 here which you can consider as constant

ishika2741 (commented):
Good Job!

ishika2741:
Good Job!

PR #1880, opened 2023-12-20
Done Design-1

m2901:


Nandha2Tech (commented):
Great Work! Kindly include TC and SC for all.
For design hash set in .js  in add method make the default value of secondary buckets to false or else it will be undefined. This will clear the issue.
add(key) {
        const primaryBucketIndex = this.primaryIndexValue(key);
        const secondaryBucketIndex = this.secondaryIndexValue(key);

        if (this.hs[primaryBucketIndex] == null) {
            if (primaryBucketIndex == 0) {
                this.hs[primaryBucketIndex] = new Array(this.secondaryItem + 1);
            } else {
                this.hs[primaryBucketIndex] = new Array(this.secondaryItem);
            }
            **this.hs[primaryBucketIndex].fill(false);**
        }
        this.hs[primaryBucketIndex][secondaryBucketIndex] = true;
    }

Nandha2Tech:
Great Work! Kindly include TC and SC for all.
For design hash set in .js in add method make the default value of secondary buckets to false or else it will be undefined. This will clear the issue.
add(key) {
const primaryBucketIndex = this.primaryIndexValue(key);
const secondaryBucketIndex = this.secondaryIndexValue(key);

    if (this.hs[primaryBucketIndex] == null) {
        if (primaryBucketIndex == 0) {
            this.hs[primaryBucketIndex] = new Array(this.secondaryItem + 1);
        } else {
            this.hs[primaryBucketIndex] = new Array(this.secondaryItem);
        }
        **this.hs[primaryBucketIndex].fill(false);**
    }
    this.hs[primaryBucketIndex][secondaryBucketIndex] = true;
}

PR #1917, opened 2023-12-31
Completed Design 1

ridashahwar:


ishika2741 (commented):
Good work! Kindly add tc and sc as well.

ishika2741:
Good work! Kindly add tc and sc as well.

PR #1921, opened 2024-01-02
add Design1'

saisurya98:


ishika2741 (commented):
Good work! Kindly add Tc and Sc as well.

ishika2741:
Good work! Kindly add Tc and Sc as well.

PR #1954, opened 2024-02-28
Done with Design-1 in python

vraj138:
Done with Design-1 in python

### DIFF
--- a/P1-HashSet.py
+++ b/P1-HashSet.py
@@ -1,0 +1,28 @@
+# Leetcode Problem No 705
+# Time Complexity : O(1)
+# Space Complexity : O(1)
+# Runtime: 107ms
+
+class MyHashSet:
+
+    def __init__(self):
+        self.myHashSet = set()
+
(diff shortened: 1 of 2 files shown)


P1-HashSet.py
```diff
@@ -0,0 +1,28 @@
+# Leetcode Problem No 705
+# Time Complexity : O(1)
+# Space Complexity : O(1)
+# Runtime: 107ms
+
+class MyHashSet:
+
+    def __init__(self):
+        self.myHashSet = set()
```
ishika2741, line 9: Do not use inbuilt set to design a hashset. Its nota good thing. Try it with the solution discussed in class.

ishika2741 (commented):
Good work!

ishika2741:
Good work!

P1-HashSet.py
vraj138: Sure. Let me do it again.

PR #1955, opened 2024-02-28
Submitting Design-1 assignments

swarnavedam:


MinStack.java
```diff
@@ -0,0 +1,44 @@
+/*
+ * Time Complexity: O(1) for all push(), pop(), top(), getMin() functions
+ * Spack Complexity: O(n)
```
ishika2741, line 3: As we discussed yesterday in class , we talk about individual function's space complexity. The space used here is in constructor , we don't take constructor's complexity

ishika2741 (commented):
Good work!

ishika2741:
Good work!

PR #2023, opened 2024-05-15
solutions to design-1

yash18082002:


Nandha2Tech (commented):
Great Work! TC for HashSet would be O(1)

Nandha2Tech:
Great Work! TC for HashSet would be O(1)

PR #2046, opened 2024-06-18
Completed Design - 1

gabbar23:
My approach was to implement a chained version of the HashSet to avoid collisions, so I utilized a List of LinkedLists. the outer list represents a collection of buckets, with each bucket being a LinkedList. When a new element is added on same index a new node of data is pushed to linkedlist.


gabbar23:
Pushed 2nd Problem of Design - 1.
Implemented 2 approaches  
      - Using single stack and Pairs data structure  
      - 2 Stacks for keeping history of new minimum found

PR #2053, opened 2024-06-19
Design-1 minstack only, py3

kashw-learns:
py3: minstack done


ishika2741 (commented):
Good work! Try to submit both the problems in same PR

ishika2741:
Good work! Try to submit both the problems in same PR

PR #2057, opened 2024-06-19
done hashset and minstack

dhikki-study:
done hashset and minstack day 1 homework


Nandha2Tech (commented):
Great Work! Don't iterate over the set instead try to check like this for contains function.
**if key in self.set1:
            return True**
Please try to implement HashSet without using builtin set() and also use double hashing technique.

Nandha2Tech:
Great Work! Don't iterate over the set instead try to check like this for contains function.
**if key in self.set1:
            return True**
Please try to implement HashSet without using builtin set() and also use double hashing technique.

PR #2100, opened 2024-07-24
done hashmap 

nehayj100:


Nandha2Tech (commented):
Kindly check the implementation of MyHashSet. The logic implemented is kind of linked list and also all the scenarios is not handled correctly. Try to use hashing to implement hashset with O(1) TC.

Nandha2Tech:
Kindly check the implementation of MyHashSet. The logic implemented is kind of linked list and also all the scenarios is not handled correctly. Try to use hashing to implement hashset with O(1) TC.

nehayj100:
Added the optimized code which works for all test cases

PR #2145, opened 2024-09-26
Completed Design 1 with solution in .py files along with Explaination at the top

Ameya-20:


Nandha2Tech (commented):

Nandha2Tech:

PR #2210, opened 2024-12-25
Completed Design-1

Chaitanya7725:


Nandha2Tech (commented):
Great Work! In design hash set, the key aspect of hash set is O(1) lookup. The approach that you used in the commented section is a linear search. To implement hash set hashing has to be there.

Nandha2Tech:
Great Work! In design hash set, the key aspect of hash set is O(1) lookup. The approach that you used in the commented section is a linear search. To implement hash set hashing has to be there.

PR #2225, opened 2024-12-27
Design-1 solutions

ccrobinson3:


Erlybird:
There is a syntax error in the code MyHashSet, python gives a decimal number if the code is a/b, to get an integer divisor you have to write a // b.

Erlybird:
There is a syntax error in the code MyHashSet, python gives a decimal number if the code is a/b, to get an integer divisor you have to write a // b.

PR #2253, opened 2025-01-22
Completed Design1

RajatOnkar:


Nandha2Tech (commented):
Great Work! In your HashSet implementation amortized time complexity is still O(1) as the secondary list size will not exceed more than 100 elements (search becomes constant after 100 elements).

Nandha2Tech:
Great Work! In your HashSet implementation amortized time complexity is still O(1) as the secondary list size will not exceed more than 100 elements (search becomes constant after 100 elements).

PR #2291, opened 2025-01-26
design1

karthikdevuluri123:
created new sample1.java file for second problem


Erlybird:
Good job! Please mention TC and SC, in design problems you can mention the TC and SC for each methods or the method that has the highest TC .

Erlybird:
Good job! Please mention TC and SC, in design problems you can mention the TC and SC for each methods or the method that has the highest TC .

PR #2327, opened 2025-03-13
python

divyanshi-srivastava:


problem1.py
```diff
@@ -0,0 +1,63 @@
+'''
+// Time Complexity : O(1)
+// Space Complexity : O(n+m)
```
ishika2741, line 3: can you please explain what is m and n here

ishika2741 (commented):
Good work!

ishika2741:
Good work!

PR #2354, opened 2025-05-09
Design - 1 PR

rutujakaushikecan:


.idea/MinStack.java
```diff
@@ -0,0 +1,42 @@
+/*
+Intuition : as this is requires min element - using PriorityQueue (min heap)
```
ishika2741, line 2: Can you try doing it without using Priority queue , that would be optimal here and also mention complexity for individual functions separately in design problems

.idea/MyHashSet.java
```diff
@@ -0,0 +1,32 @@
+/*
+Intuition : as this is limited number  (10 ^ 6), we can use boolean arrays
```
ishika2741, line 2: This is not the optimal solution. You created an array of 1M buckets which is wastage of space.

ishika2741 (commented):
Please try to submit optimal solutions for both the problems.

ishika2741:
Please try to submit optimal solutions for both the problems.

rutujakaushikecan:
@ishika2741 I updated optimal solutions, please review

PR #2441, opened 2025-06-18
Completed Design-1

rk-peddibhotla:


super30admin:
Strengths:
- The solution correctly implements all required operations with the expected time complexity.
- The use of two stacks to track the minimum values is a valid approach.

Areas for Improvement:
- The pop operation could be optimized by only updating the min when the popped element is the current min. This would reduce unnecessary updates to the min variable.
- The edge case handling in getMin could be more robust. Currently, it prints a message and returns -1, but it might be better to throw an exception or return a more meaningful value.
- The main method is included for testing, which is good, but it could be moved to a separate test class for better separation of concerns.

PR #2462, opened 2025-06-20
Pr design1 -done

sindhoorpalla:


super30admin:
- MinStack:
  - The `push()` method should only push the old minimum when the new value is less than or equal to the current minimum.
  - The `pop()` method should check if the popped value is the current minimum and restore the previous minimum if necessary.
  - The `getMin()` method should always return the correct minimum value, even after pops.
  
- MyHashSet:
  - The `add()` method should check if `bucketItem` is within bounds before accessing the array to avoid potential ArrayIndexOutOfBoundsException.
  - Consider initializing the secondary array with a fixed size to prevent dynamic resizing issues.

PR #2474, opened 2025-06-23
added solutions for DesignHashSet and MinStack

sriram595:


super30admin:
Strengths:
- The MyHashSet implementation correctly handles the edge case for the 0th bucket, showing good attention to detail.
- The MinStack implementation is elegant and efficient, using a single stack to maintain both values and the minimum.
- Code is well-commented and follows good naming conventions.

Areas for Improvement:
- For MyHashSet, consider adding comments to explain why the 0th bucket needs an extra slot (to handle the key 1,000,000).
- In MinStack, while the implementation is correct, it might be worth noting in comments why we push the old minimum before pushing the new value (to restore it when the new minimum is popped).

PR #2476, opened 2025-06-24
completed Design1

ramya-gopalaswamy:


super30admin:
Strengths:
- The implementation of both problems is correct and efficient.
- Proper memory management is handled in MyHashSet with the destructor.

Areas for Improvement:
- In MyHashSet, consider using smart pointers (e.g., std::unique_ptr) to manage the secondary vectors automatically, reducing the risk of memory leaks.
- For MinStack, the initial minVal could be initialized directly in the constructor initializer list for better style.

PR #2490, opened 2025-07-24
Done Design-1

akhilsai25:


super30admin:
Strengths:
- The MinStack implementation is clean and efficient, with clear comments explaining the logic.
- The MyHashSet implementation correctly handles collisions using chaining and includes dummy nodes to simplify edge cases.
- Both solutions are well-commented and easy to follow.

Areas for Improvement:
- For MyHashSet, consider using a more balanced data structure for secondary storage to ensure O(1) time complexity in all cases.
- The getPrev method in MyHashSet could be optimized to stop early if the key is found, though this is a minor improvement.
- The MinStack's top method returns -1 when the stack is empty, which might not be the expected behavior (throwing an exception might be more appropriate).

PR #2496, opened 2025-07-26
Done Design-1

Lavanyarekkala:
Completed Design-1

### DIFF
--- a/MinStack.java
+++ b/MinStack.java
@@ -1,0 +1,72 @@
+// Time Complexity : O(1)
+// Space Complexity :O(1)
+// Did this code successfully run on Leetcode : Yes
+// Any problem you faced while coding this : No
+
+
+
+// Your code here along with comments explaining your approach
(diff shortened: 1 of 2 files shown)


super30admin:
Strengths:
- The student has correctly implemented both data structures with the required time and space complexity.
- The use of comments to explain the approach is excellent and helps in understanding the code.
- The double hashing implementation in MyHashSet is well-done and handles edge cases like the primary bucket being 0.

Areas for Improvement:
- In MinStack, the check `if(minSt==null)` is redundant since `minSt` is initialized in the constructor and never set to null.
- Naming conventions could be more consistent. For example, method names like `GetPrimaryHashKey` should follow camelCase (e.g., `getPrimaryHashKey`) to align with Java conventions.
- The MinStack implementation could be simplified by always pushing the current minimum value before pushing the new value, which would make the pop operation more straightforward.

PR #2502, opened 2025-07-28
Completed Design 1: Design HashSet

NehaKhan9753:


super30admin:
- The student has done an excellent job implementing the HashSet from scratch using double hashing and lazy initialization. The code is clean, efficient, and correctly handles all operations.
- The comment about space complexity being O(N) could be clarified. Since the primary and secondary buckets are fixed in size (1000 and 1000+1), the space complexity is actually O(1) regardless of the number of keys.
- The student could consider adding a brief explanation of why the secondary bucket size is increased by 1 when the primary index is 0 (to handle the edge case for key = 1,000,000).
- Overall, the solution is well-written and demonstrates a good understanding of the problem and the data structure.

PR #2514, opened 2025-08-20
hashSet, minStack Done

hp6318:


super30admin:
Strengths:
- Edge cases are handled explicitly, which is a good practice.

Areas for Improvement:
- In MyHashSet, the handling of the last bucket (index 1000) could be simplified. Instead of treating it as a special case, consider using a uniform approach for all buckets.
- In MinStack, setting `curr_min` to `float('inf')` initially would make the code more intuitive and easier to understand.
- The comments could be more detailed to explain why certain design choices were made, such as the choice of 1001 buckets in MyHashSet.

PR #2531, opened 2025-08-22
Completed both the problems

restacksyj:


super30admin:
Strengths:
- The use of double hashing in MyHashSet is a good approach to handle collisions.
- The comments and explanations provided are clear and helpful.

Areas for Improvement:
- Avoid using `#include <bits/stdc++.h>` in C++ code. Instead, include only the necessary headers (e.g., `<vector>`, `<stack>`).
- Consider adding more edge case tests to ensure robustness, especially for MyHashSet (e.g., testing with keys at the boundaries like 0 and 1e6).
- For MinStack, the student could consider using a single stack that stores pairs of values and the current minimum, which might be more memory efficient in some cases.

PR #2571, opened 2025-11-12
Anil my hashset.java

anilkumarm0921-oss:
image
image


super30admin:
Strengths:
- Correct implementation of both problems with appropriate time and space complexity.
- Good handling of edge cases in the hash set implementation.
- Clear separation of concerns in the MinStack implementation.

Areas for Improvement:
- Improve variable naming for better readability (e.g., 'a' and 'b' in MinStack).
- Move comments from Sample.java to the relevant files where they would be more useful.
- Consider simplifying the hash set implementation by always using a secondary bucket size of 1001 to avoid the special case for index 0.
- Add more detailed comments explaining the design choices, especially for the hash set's double hashing approach.

PR #2572, opened 2025-11-12
Day2- design-hashset, min-stack

tanvipandhre:


super30admin:
Strengths:
- Correct implementation of both problems with optimal time and space complexity.
- Good use of helper methods (hash1, hash2) to improve code readability.
- Clear and concise README explaining the approach.

Areas for Improvement:
- Consider adding inline comments in the hashset implementation to explain why bucket 0 needs an extra element, as this is a non-obvious edge case.
- In the min stack implementation, the `min` variable is redundant since `minSt.peek()` always gives the current minimum. You could remove `min` to simplify the code slightly.

PR #2573, opened 2025-11-13
Completed the Design1

SaiSaranPathuriNEU:


SaiSaranPathuriNEU:
Please kindly observe that code comments(approach explaination) are baked in the code itself

SaiSaranPathuriNEU:
Also the naming convention is not upto the mark, next assignment i'll make sure i'll be bit more clear. I was timing how much time i take to solve these, so focused on explaination as comments and speed. I'll improve the naming for next assignements. Thank you!

super30admin:
Strengths:
- The student has correctly implemented the required functionalities for both problems.
- The use of separate hash functions and initialization in the constructor shows good practice.

Areas for Improvement:
- For the min-stack, consider optimizing the space usage by only pushing to the min stack when a new minimum is found, rather than pushing the current minimum every time. This would reduce the space usage in some cases.
- The comment in the hashset solution about "dinamically allocate" should be corrected to "dynamically allocate."
- The space complexity comment in the hashset solution could be more precise, mentioning that it's O(n) in the worst case but typically much less due to the hashing.

PR #2594, opened 2025-12-24
Design 1 completed

thondaha:
Completed HomeWork 1 and 2


super30admin:
Strengths:
- The implementations are correct and handle edge cases well.
- The code is clean, well-commented, and follows good practices.
- The use of separate chaining in MyHashSet and the minSoFar approach in MinStack are well thought out.

Areas for Improvement:
- Consider using a prime number for the bucket size in MyHashSet to potentially reduce collisions.
- The MinStack implementation could include null checks in pop(), top(), and getMin() to handle cases where the stack is empty, though the problem constraints might not require it.
- Adding more test cases in the main method to cover edge cases would be beneficial.

PR #2601, opened 2025-12-26
Queue design using two stacks

dbkuppagiri:


super30admin:
- The solution is well-implemented and correctly solves the problem of implementing a queue using stacks.
- The time complexity analysis should be mentioned in the comments to help others understand the performance characteristics of each operation.
- The peek method could handle the case where both stacks are empty more gracefully (currently it returns undefined, which might not be the expected behavior in all contexts).
- Consider adding a check in the pop method to handle the case where the queue is empty (currently it would return undefined without any indication of an error).

Overall, the solution is good, but it could be improved with better error handling and more detailed comments about time complexity.

PR #2605, opened 2025-12-27
My_design_hashset

SrinivasaSupreeth:


super30admin:
Strengths:
- Your code is clean, concise, and easy to understand.
- You've correctly implemented all required operations with proper method signatures.
- The use of Python's built-in set ensures good performance characteristics.

Areas for Improvement:
- The problem likely expects you to implement a hash set from scratch rather than using Python's built-in set. This would demonstrate your understanding of hashing and collision resolution techniques.
- Consider implementing your own hashing mechanism and handling collisions (e.g., using chaining or open addressing).
- The solution doesn't show your understanding of the underlying data structure principles that the problem is likely testing.

For a more complete solution, you might want to:
1. Choose an appropriate size for your hash table
2. Implement a hash function
3. Handle collisions (perhaps using chaining with linked lists)
4. Implement resizing logic if needed

PR #2610, opened 2025-12-28
Completed Design 1(HashSet and minStack

Pranjal0211:


super30admin:
Strengths:
- Good understanding and implementation of double hashing technique.
- Proper handling of edge cases (like when primaryIndex=0).
- Clear and concise comments explaining the approach.
- Correct use of lazy initialization for secondary arrays to save space.

Areas for Improvement:
- There's a typo in the variable name `secondaryBucekts` (should be `secondaryBuckets` for consistency).
- The comment about getting confused about primary index being null could be addressed by better understanding that null means no secondary array has been initialized for that primary index yet.
- The initialization of `storage` in the constructor uses a hardcoded 1000 instead of `primaryBuckets`, which could lead to inconsistencies if `primaryBuckets` is changed later.

Pranjal0211:
@super30admin Hello, I submitted the Design1 Part 2 MinStack now. Just commenting to keep in loop.

PR #2617, opened 2026-01-12
done

ShrinidhiMane:


super30admin:
Strengths:
- The MinStack implementation is correct and efficient, using the two-stack approach appropriately.
- The code is readable and well-structured in both implementations.
- Good use of Java's built-in data structures where appropriate (for MinStack).

Areas for Improvement:
- For MyHashSet, you should implement the HashSet functionality from scratch rather than using Java's built-in HashSet. The problem likely expects you to demonstrate understanding of how a HashSet works internally.
- The return statements in MyHashSet's add and remove methods are unnecessary since these methods are void.
- The contains method in MyHashSet can be simplified to just `return s.contains(key);` without the if-else.

PR #2623, opened 2026-02-03
2 problems completed

BharathVuppala96:


super30admin:
Your solution is good and follows the double hashing approach correctly. However, I noticed that in your `hash1` and `hash2` functions, you are using hard-coded 1000 instead of using the instance variables `self.bucket` and `self.bucketItem`. This is a problem because if you ever change the value of `bucket` or `bucketItem`, the hash functions won't reflect that change. You should use the instance variables to ensure consistency.

For example, instead of:
```python
def hash1(self,key:int)->int:
    return key%1000
```
you should write:
```python
def hash1(self,key:int)->int:
    return key % self.bucket
```

Similarly for `hash2`:
```python
def hash2(self,key:int)->int:
    return key // self.bucketItem
```

This change will make your code more maintainable and prevent potential bugs.

Another minor point: in the `add` method, you have:
```python
if self.storage[bucket]==None:
    if bucket==0:
        self.storage[bucket]=[False]*(self.bucketItem+1)
    else:
        self.storage[bucket]=[False]*(self.bucketItem)
```
This is correct. However, note that in the reference solution, the secondary array for bucket 0 is of size `secondaryBuckets+1` (which is 1001) to accommodate the key 1000000. Since your `bucketItem` is 1000, key 1000000 would have primary index 0 and secondary index 1000. Without the +1, the secondary array for bucket 0 would have indices 0 to 999, so index 1000 would be out of bounds. So your handling for bucket 0 is correct.

Overall, your solution is correct and efficient. With the small change to use instance variables in the hash functions, it would be excellent.

PR #2625, opened 2026-02-03
Hashset implemantation

takchiks:


super30admin:
Strengths:
- You have chosen a large prime number for HASH_SIZE, which helps in distributing keys uniformly.
- The code is structured and easy to read.
- You have considered the constraints and allocated sufficient space.

Areas for Improvement:
1. The removal method is incorrect. When you remove a key by setting it to -1 and decrementing the count, the array now has a "hole" and the linear scans only go up to the count, so the -1 value is not checked in future contains calls. However, if you add the same key again, it might be added twice if there is a hole? Actually, the add method checks for existence only up to the count, so it might not see the -1 value. This leads to duplicates and incorrect state. Instead, you should overwrite the removed element with the last element in the bucket to maintain a contiguous array. For example:
   ```
   for (int i = 0; i < bucketCount[h]; i++) {
        if (buckets[h][i] == key) {
            buckets[h][i] = buckets[h][bucketCount[h]-1];
            bucketCount[h]--;
            return;
        }
   }
   ```
   This way, the array remains contiguous without holes.

2. The add method does not handle the case when the bucket is full. Although the bucket size is large enough to avoid collisions in most cases, it is possible (though unlikely) for a bucket to exceed BUCKET_SIZE. Since the problem states at most 10^4 calls, and HASH_SIZE is 78727, the average chain length is very low. But to be safe, you should consider dynamic resizing or a different data structure for buckets (like ArrayList). However, given the constraints, it might not be necessary. But if you want to be correct, you should ensure that no bucket overflows.

3. Alternatively, you can use a simpler approach: since the key range is only 0 to 10^6, you can use a boolean array of size 1000001. This would use about 1e6 booleans (1 MB) and all operations would be O(1). This is simpler and efficient.

4. The current solution uses more memory than necessary. Consider using a more memory-efficient design.

super30admin:
Strengths:
- You used a large prime number for the hash size, which is good for distributing keys.
- The code is structured and variables are named clearly.
- You considered the constraints and allocated sufficient space.

Areas for Improvement:
1. The removal method is incorrect. Instead of setting the key to -1 and decrementing the count, you should overwrite the removed element with the last element in the bucket (if you want to maintain a list without holes) or use a flag to mark removed elements. However, the current approach breaks the linear scan because the bucketCount reduces, so the removed element is no longer scanned, but the hole remains. Actually, in your current code, after removal, the element is set to -1 and the count is reduced, so the next scan will not see it. But the problem is that the bucket array now has a hole that is not used. When adding, you only check up to the current count, so you might not reuse the hole. This leads to wasted space and potential bucket overflow even if there are empty slots.

   Suggestion: In the remove method, you can swap the removed element with the last element in the bucket and then decrement the count. This way, the bucket remains contiguous. For example:
   ```
   for (int i = 0; i < bucketCount[h]; i++) {
        if (buckets[h][i] == key) {
            buckets[h][i] = buckets[h][bucketCount[h]-1];
            bucketCount[h]--;
            return;
        }
   }
   ```
   But note: this changes the order, but that is acceptable.

2. The add method does not handle the case when the bucket is full. Although the chosen parameters make overflow unlikely, it is possible. Since the problem states at most 10^4 calls, and your bucket size is 70, and hash size is 78727, the load factor is very low. However, to be safe, you should consider dynamic resizing or a better collision resolution method. Alternatively, you can use a linked list for each bucket to avoid fixed size.

3. The code includes an unrelated MinStack class. This should be removed to avoid confusion.

4. The contains method should also skip negative values if you are using -1 for removed, but in your current implementation, you don't need to because you reduce the count. However, if you change to reuse holes, you might need to handle negative values.

5. Consider using a more standard approach like the reference solution, which uses a 2D boolean array with double hashing. This avoids the need for storing integers and uses less space.

PR #2631, opened 2026-02-05
Create hashset.py

chvsnagarjun:


super30admin:
Your solution is well-implemented and follows the double hashing technique correctly. Here are some points to consider:

1. **Correctness**: Your solution should work correctly for all keys in the range [0, 10^6]. However, note that when you initialize the secondary array for the first bucket (index 0), you use `self.size + 1` to account for the key 1000000. This is correct because 1000000 % 1000 = 0 and 1000000 // 1000 = 1000, so without the +1, the index would be out of bounds. Good job on handling that edge case.

2. **Variable Names**: While `hash1` and `hash2` are clear, consider using more descriptive names like `primary_index` and `secondary_index` to make the code even more readable.

3. **Comments**: The comments at the top are a good start, but they could be more specific. For example, you might explain why you chose the size 1000 and how the double hashing works.

4. **Efficiency**: Your solution is efficient. However, in the `remove` method, you set the value to `False`. This is correct, but note that you never deallocate the secondary array even if all values in it become false. This is acceptable because the problem does not require that, and it might be inefficient to constantly allocate and deallocate. But if memory usage is a concern, you might consider deallocating the secondary array if it becomes entirely false. However, given the constraints, it's not necessary.

5. **Edge Cases**: Consider what happens if the key is negative? The problem states that key is between 0 and 10^6, so negative keys are not expected. But if the problem were to change, your code might break. However, since the constraints are clear, it's not an issue here.

6. **Python Specifics**: In Python, using a list of lists of booleans is a valid approach. However, you might also consider using a list of sets or other structures for better memory efficiency if the keys are sparse. But for this problem, your solution is appropriate.

super30admin:
Your solution is well-structured and follows the double has

PR #2632, opened 2026-02-05
S30 | FAANMG Problem #2 | design-hashset

sandy7907:


super30admin:
Strengths:
- You have chosen a good approach using a two-dimensional boolean array to represent the hash set.
- The code is well-organized and easy to read.
- You have implemented the basic structure of the methods correctly.

Areas for Improvement:
- The primary hashing function is incorrect. Instead of `key / primaryArraySize + 1`, you should use modulo arithmetic for the primary hash. The reference solution uses `key % primaryBuckets` for the primary index. This ensures a uniform distribution and avoids out-of-bound indices.
- The secondary hashing function should be `key / secondaryBuckets` as in the reference solution, not `key % secondaryArraySize`. This is because for the secondary array, you need to map the key to a position within the secondary bucket. The reference solution uses integer division to get the secondary index.
- You need to handle the edge case for the first primary bucket (index 0) to accommodate the maximum key (1000000). In the reference solution, when the primary index is 0, the secondary array is allocated with size 1001 instead of 1000. This is because for key=1000000, the primary index is 0 (since 1000000 % 1000 = 0) and the secondary index is 1000000 / 1000 = 1000, which requires an array of size 1001. Without this, an ArrayIndexOutOfBoundsException would occur.
- Consider using more conventional method names like `getPrimaryHash` and `getSecondaryHash` for brevity.

To correct your solution:
- Change `getPrimaryHashValue` to `return key % primaryArraySize;` (but note that your primaryArraySize is 1001, which is different from the reference's 1000). Alternatively, you can set primaryArraySize to 1000 and then handle the edge case for index 0.
- Change `getSecondaryHashValue` to `return key / secondaryArraySize;` (and set secondaryArraySize to 1000).
- In the `add` method, when allocating the secondary array for a primary index, if the primary index is 0, you should allocate an array of size 1001 (if secondaryArraySize is 1000) to avoid index out of bounds for key=1000000.

Example correction:
```
public class MyHashSet {
    int primaryArraySize = 1000;
    int secondaryArraySize = 1000;
    boolean[][] customHashArray;

    public MyHashSet() {
        this.customHashArray = new boolean[primaryArraySize][];
    }
    
    private int getPrimaryHash(int key) {
        return key % primaryArraySize;
    }

    private int getSecondaryHash(int key) {
        return key / secondaryArraySize;
    }

    public void add(int key) {
        int primaryPos = getPrimaryHash(key);
        int secondaryPos = getSecondaryHash(key);
        if (customHashArray[primaryPos] == null) {
            if (primaryPos == 0) {
                customHashArray[primaryPos] = new boolean[secondaryArraySize + 1];
            } else {
                customHashArray[primaryPos] = new boolean[secondaryArraySize];
            }
        }
        customHashArray[primaryPos][secondaryPos] = true;
    }
    
    // remove and contains similarly adjusted
}
```

super30admin:
Strengths:
- The code is clean and well-organized with meaningful variable names.
- The hashing logic is correctly implemented for the primary and secondary functions.
- The methods are concise and follow the expected behavior for the hash set operations.

Areas for Improvement:
- The initialization of the secondary array for primary index 0 is flawed. The condition `if(primaryPos == 0 && secondaryPos == 1000)` is incorrect because it checks the current key's secondary index to decide the size. However, the secondary array should be allocated with enough size to handle all possible keys that hash to that primary bucket. For primary index 0, the keys can range from 0 to 1000000, which requires a secondary array of size 1001 (since secondary index for 1000000 is 1000). The reference solution handles this by always allocating 1001 elements for primary index 0. You should change the initialization to:
  ```
  if (primaryPos == 0) {
      customHashArray[primaryPos] = new boolean[secondaryArraySize + 1];
  } else {
      customHashArray[primaryPos] = new boolean[secondaryArraySize];
  }
  ```
  This ensures that for primary index 0, the secondary array has indices 0 to 1000 (total 1001), and for others, it has 1000 indices (0 to 999).

- Additionally, in the `remove` method, you should check if the secondary array is null before accessing it. Currently, you check `if(customHashArray[primaryPos] != null)`, which is correct, but you should also ensure that the secondary index is within bounds. However, since the secondary array is allocated only when needed and the hashing functions are consistent, it should be safe. But the initialization fix above will prevent any index out of bounds.

- It is good practice to include comments to explain the reasoning behind certain decisions, such as why primaryArraySize and secondaryArraySize are chosen as 1000, and why we need an extra slot for primary index 0.

Overall, the solution is very close to being correct. With a small fix in the initialization logic, it will work perfectly.

PR #2635, opened 2026-02-07
S30 FAANMG Problem #4 | Create Queue using Stack

sandy7907:


super30admin:
Strengths:
- The code for MyHashSet follows the double hashing approach correctly in general.
- Variable names are descriptive (e.g., primaryArraySize, secondaryArraySize).
- The methods are well-structured and easy to read.

Areas for Improvement:
- The condition in the add method for allocating the secondary array for primary index 0 is incorrect. It should only check if primaryIndex is 0, and then allocate an array of size secondaryArraySize + 1 to accommodate the key 10^6 (which hashes to primary 0 and secondary 1000). The current condition (primaryPos==0 && secondaryPos==1000) will only allocate the larger array when adding key 1000000, but for other keys in primary 0 it will allocate the normal size. This means that when adding key 1000000, the secondary array might not have been allocated with enough size, and when adding other keys in primary 0, the array might be too small if the key requires index 1000. You should mimic the reference solution: for primary index 0, always allocate an array of size secondaryBuckets+1, and for others, allocate secondaryBuckets.
- The MyQueue class is included, which is not part of the problem. Ensure that you only submit the code required for the problem.

sandy7907:
MinStack is the new problem that have been updated as part of Design 1. Please let me know if I need to close this PR delete the old problem and only push MinStack

super30admin:
Strengths:
- The code is well-organized with clear separation of methods.
- Variable names are descriptive (e.g., primaryArraySize, secondaryArraySize, customHashArray).
- The hashing functions are correctly implemented to distribute keys.

Areas for Improvement:
1. The condition in the add method for creating the secondary array for primary index 0 is flawed. Currently, it checks `if(primaryPos == 0 && secondaryPos == 1000)`, which means it only creates the larger array when you are adding the key 1000000. However, for primary index 0, you need a secondary array of size 1001 to accommodate keys from 0 to 1000000 (which maps to secondary indices 0 to 1000). The condition should be similar to the reference solution: if primary index is 0, create an array of size secondaryArraySize+1. This should be done regardless of the key being added at that moment, because any key with primary index 0 might be up to 1000000.

   Correction:
   Instead of:
   ```java
   if(primaryPos == 0 && secondaryPos == 1000) {
        customHashArray[primaryPos] = new boolean[secondaryArraySize + 1];
   } else {
        customHashArray[primaryPos] = new boolean[secondaryArraySize];
   }
   ```
   It should be:
   ```java
   if (primaryPos == 0) {
        customHashArray[primaryPos] = new boolean[secondaryArraySize + 1];
   } else {
        customHashArray[primaryPos] = new boolean[secondaryArraySize];
   }
   ```

2. The MinStack class is included in the same file, which is not required for this problem. It is important to submit only the code relevant to the problem. If this was a mistake in the submission, please ensure to only include the MyHashSet class.

3. Although not critical, you can avoid recalculating the secondary hash in remove and contains by storing it in a variable when needed. However, the current way is acceptable.

Overall, the solution is close to correct, but the handling of the edge case (key=1000000) is not robust. With the correction, it will work correctly.

PR #2658, opened 2026-04-15
Design-1 homework

richapatil:


super30admin:
### Implement Hash Set (705. Design HashSet/MyHashSet.java)
Strengths:
- The solution correctly implements the double hashing technique with appropriate primary and secondary hash functions.
- The edge case for primary index 0 is handled correctly by allocating an array of size secondaryBucket + 1.
- The code is well-structured and readable with clear variable names and methods.

Areas for improvement:
- The method names `primaryKey` and `secondaryKey` might be slightly misleading because they return indices, not keys. It would be more accurate to name them `getPrimaryIndex` and `getSecondaryIndex` or similar to reflect that they are hash functions computing indices.
- There is a minor inconsistency: the variable `secondaryBucket` is used in the same way as `secondaryBuckets` in the reference solution, but the reference solution uses a different variable name for the secondary bucket size. However, this is just a naming choice and does not affect functionality.
- The code could include comments to explain why the extra slot is needed for primary index 0, which might help others understand the design.

Overall, the solution is correct and efficient.

VERDICT: PASS

---

### Implement Min Stack (155. Min Stack/MinStack.java)
Strengths:
- The solution is correct and efficient, with O(1) time for each operation and O(n) space.
- The code is well-structured and readable, with clear variable names and comments.
- The initialization of minst with Integer.MAX_VALUE is handled correctly.

Areas for improvement:
- In the push method, the condition `if(val < min)` is correct, but note that the reference solution uses `min = Math.min(val, min);` which is more concise and achieves the same. However, the student's approach is also valid.
- The comment "setting minStack value to infinity" is slightly misleading because Integer.MAX_VALUE is not exactly infinity, but it is a common convention to represent a large value. It might be better to say "initializing with a large value".
- The comment "TC - O(1)" and "SC - O(n)" is good, but it should be noted that the space complexity is O(n) due to the two stacks, which is optimal for this approach.

Overall, the student has a good understanding of the problem and has implemented a correct solution.

VERDICT: PASS

PR #2663, opened 2026-05-15
Design-1 problems solution

Simranb10:
Submit Design-1 problem solutions


super30admin:
### Implement Hash Set (MyHashMap.java)
**Strengths:**
- Good use of a separate Node inner class for encapsulation
- Reasonable variable naming and structure
- Attempted to handle collisions through chaining

**Areas for Improvement:**
- **Critical**: You implemented a HashMap instead of a HashSet. The problem requires storing only keys (no values), with methods `add()`, `contains()`, and `remove()`. Your implementation has `put()` and `get()` which are HashMap methods.
- The linked list approach is more complex than necessary. A simple boolean array approach (like the reference solution) would be more efficient and easier to implement.
- The `getPrevious()` method has convoluted logic that makes the code harder to follow.
- Consider using a boolean array instead of linked lists for O(1) guaranteed operations.

VERDICT: NEEDS_IMPROVEMENT

---

**Strengths:**
- Good attempt at optimizing space by only tracking when minimum changes
- Clean variable naming and reasonable structure
- Correctly identifies O(1) time complexity

**Areas for Improvement:**
1. **Critical Bug**: The `top()` method returns incorrect values. When you push the previous minimum before the new value, the stack structure becomes corrupted for top operations.
2. **Logic Flaw**: The push logic `if(min >= val)` causes duplicate minimums to be pushed, which breaks the correspondence between the main stack and minimum tracking.
3. **Consider using a two-stack approach**: The reference solution uses a separate `minSt` stack that mirrors every push/pop operation, making the code more predictable and easier to debug.
4. **Alternative fix**: If keeping single stack, you need to ensure that `top()` always returns the actual top element, not a minimum value.

The core issue is that mixing data elements with minimum tracking elements in a single stack creates ambiguity about what `top()` should return.

VERDICT: NEEDS_IMPROVEMENT

